PR #30, opened 2017-05-04
Changed models to match the merged db created by 'ilikecats'

Lietzu:
ilikecats db is here: https://kitsunemimi.pw/tmp/merged.sqlite3.zip

Biggest change was creating a separate table for statuses.

There's some small changes in other places to make everything work together too.

### DIFF
--- a/models.go
+++ b/models.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"github.com/jinzhu/gorm"
 	"html"
 	"html/template"
 	"strconv"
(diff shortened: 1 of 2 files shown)


models.go
```diff
@@ -19,18 +18,28 @@ type Categories struct {
...
 	Id              int            `gorm:"column:torrent_id"`
 	Name            string         `gorm:"column:torrent_name"`
 	Category_id     int            `gorm:"column:category_id"`
 	Sub_category_id int            `gorm:"column:sub_category_id"`
-	Status          int            `gorm:"column:status_id"`
+	Status_id	int	       `gorm:"column:status_id"`
 	Hash            string         `gorm:"column:torrent_hash"`
+	Date		int	       `gorm:"column:date"`
+	Downloads	int	       `gorm:"column:downloads"`
+	Filesize	string	       `gorm:"column:filesize"`
```
avail: should be converted to bytes and stored as integer

Lietzu: Unfortunately the current db has them all stored as text, at some point that will all have to be converted for proper sorting.

PR #34, opened 2017-05-05
Script for binary package cross-compilation

bakape:
Should easy distribution. Works from Linux and should also work from a mingw64 shell on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@
 *.db
 main
 nyaa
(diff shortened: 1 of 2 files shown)


avail:
building works fine, but 'zip' isn't a valid command on windows :(

bakape:
You can install it in MSYS2. I forgot raw mingw does not have a package
manager.

PR #60, opened 2017-05-05
Rearranged Files

akuma06:
Configurations are separated in files in the folder config
Connection to database by a package to import when needed
Models will be in a model package for better maintenance
Services access to the models
Utils are tools or functions that can be used anywhere
main.go cleaned a bit and other files modifications are there for the above modifications


ewhal:
@akuma06 https://github.com/ewhal/nyaa/invitations

PR #63, opened 2017-05-05
Remake search is now Filter Remakes like old nyaa

Lietzu:
Old nyaa had Filter Remakes as option 2 on the drop down menu, since they were generally considered undesirable. 

Imo there should maybe both options, but trying to make it close to OG nyaa as possible as of right now.

### DIFF
--- a/main.go
+++ b/main.go
@@ -191,7 +191,11 @@ func searchByQuery(r *http.Request, pagenum int) (SearchParam, []Torrents, int) 
 		parameters.params = append(parameters.params, searchSubCatId)
 	}
 	if search_param.Status != "" {
-		conditions = append(conditions, "status_id = ?")
+		if search_param.Status == "2" {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -158,8 +158,12 @@ func searchByQuery(r *http.Request, pagenum int) (SearchParam, []model.Torrents,
...
 	}
 	if search_param.Status != "" {
-		conditions = append(conditions, "status_id = ?")
-		parameters.Params = append(parameters.Params, search_param.Status)
+		if search_param.Status == "2" {
+			conditions = append(conditions, "status_id != ?")
+		} else {
+			conditions = append(conditions, "status_id = ?")
+		}
+		parameters.params = append(parameters.params, search_param.Status)
```
sfan5: you accidentally lowercased `Params`

Lietzu:
Fixed.

PR #72, opened 2017-05-05
RSS feed with common search

sazuna:
Pretty straightforward, just generating the RSS feed based on the current search parameters.


akuma06:
Already put it in the commit #40cef11

akuma06:
But your approach might be better. If there is a member section, the way I've generated the link would add unuseful query parameters in the url

PR #76, opened 2017-05-06
Add Ren-chon to corner.

chuu2b:


aerojun:
Shouldn't it also be the favicon?

chuu2b:
Right you are.

PR #84, opened 2017-05-06
Adding Docker support (WIP)

tomleb:
# WIP
Don't merge yet, just putting it out there while I'm working on it.

The sqlite backend has been tested and is working.

The postgres can't be tested yet because the postgres dump we have is not compatible with the application. Whoever is working on the migration (ilikecats I believe) please contact me when you have a dump that is compatible.

I think we also wanted to have mysql so I'll add that, but again we need to have databases that are compatible. (Or I could just build a fake db in the meantime)


akuma06:
I'm here to rescue you !

deploy/init.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -eux
+
+# TODO Doesn't scale, find another way to wait until db is ready
+if [[ "${PANTSU_DBTYPE}" = "postgres" ]]; then
+  echo 'Waiting for the database to be ready...'
+  sleep 40
+fi
+
+go get github.com/ewhal/nyaa
```
sfan5: missing pull of `qtc` binary?

tomleb: Guess we won't need it for long with the revert on templates, should I add it or wait for the revert PR to be merged ?

tomleb:
The docker-compose.postgres.yml file assume we want to put postgres in a container. I'd prefer not to put the database in a container for production but I want to keep it for dev. Should I create another file say docker-compose.postgres-prod.yml ?

tomleb:
Added production file for when we deploy without db inside docker.

This PR is ready to be merged, but I'd still like people to test it before..

PR #89, opened 2017-05-06
`filesize` as INTEGER

sfan5:
requires v6 db

### DIFF
--- a/util/format.go
+++ b/util/format.go
@@ -1,0 +1,24 @@
+package util
+
+import (
+    "fmt"
+)
+
+func FormatFilesize(bytes int64) string {
+    var unit string
+    var value float64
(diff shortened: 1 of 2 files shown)


util/format.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+    "fmt"
+)
+
+func FormatFilesize(bytes int64) string {
```
nitori: There are also some files that have "TiB". I checked on sqlite3 db with:

```
SELECT
  SUBSTR(filesize, -4),
  COUNT(SUBSTR(filesize, -4)) AS nums
FROM torrents
GROUP BY SUBSTR(filesize, -4)
ORDER BY nums DESC;
```

PR #96, opened 2017-05-06
Models for User, Role and Comment  …

akuma06:
modelHelper for getting value from request and apply it to a Form
Getting value from a Form and apply to a Model


majestrate:
try running `go fmt`

akuma06:
FINALLY !!!! :p

PR #98, opened 2017-05-06
Don't reparse templates on every fucking request

bakape:


sfan5:
This means no more "hot-reload" like functionality for static files, correct?

bakape:
I can implement hot reload through a file watcher, if you really need it. Reparsing templates on each request is simply very dumb. Might as well be using PHP at that point.

PR #104, opened 2017-05-06
Added tool for sending emails + User Service (WIP)

akuma06:


majestrate:
does gomail do dkim? if not then the outgoing mails are going to go straight to the spam folder.

akuma06:
Dont know but you can use another smtp server than your own. Just have to put it in config "email.go" file

PR #106, opened 2017-05-06
Change Fap to the end of the navbar

idontusegithubkek:
It just doesn't feel right for Fap to be anywhere besides the rightmost position


ayame-git:
>muh legacy

PR #115, opened 2017-05-06
Drop "Category", "Sub_Category" and "Statuses" tables

sfan5:
(also includes rly important change to faq)


yiiTT:
On the topic of "pirateparty.gr tracker is the same as the coppersurfer one" it is always good to have another domain of the same tracker.

PR #117, opened 2017-05-06
Add subcategory name to torrent view page

kipukun:
This more closely resembles the original nyaa view page, and it also saves us an image call.
capture


kipukun:
As I made this PR, someone else did the same thing. Closing.

PR #123, opened 2017-05-06
Revert quicktemlate search template

bakape:
Hope nothing forgotten.


tomleb:
The docker-compose.postgres.yml file assume we want to put postgres in a container. I'd prefer not to put the database in a container for production but I want to keep it for dev. Should I create another file say docker-compose.postgres-prod.yml ?

PR #127, opened 2017-05-07
Captchas

bakape:
Just so someone does not needlessly duplicate the work on this. Will try to finishing tomorrow.


akuma06:
Would it not be better and more secure to use recaptcha ?

bakape:
Yes, but that would: 
- expose information to Google, which many of the core users are vary of
- needlessly complicate mirroring by requiring to register with Google and setup ReCaptcha keys
- add an extra dependency on an external service

service/captcha/capthca.go
```diff
@@ -0,0 +1,97 @@
+package captcha
```
Atvaark: There's a typo in these filenames:

- templates/_capthca.html
- service/captcha/capthca.go

PR #131, opened 2017-05-07
Implement upload for new torrents

azhao12345:
we need some form of registration or antispam before we can enable uploads though

wip

### DIFF
--- a/router/uploadHandler.go
+++ b/router/uploadHandler.go
@@ -1,10 +1,14 @@
 package router
 
 import (
+	"fmt"
+	"github.com/ewhal/nyaa/db"
+	"github.com/ewhal/nyaa/model"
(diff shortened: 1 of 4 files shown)


router/uploadHandler.go
```diff
@@ -22,7 +26,24 @@ func UploadHandler(w http.ResponseWriter, r *http.Request) {
 		if err == nil {
 			//validate name + hash
 			//add to db and redirect depending on result
+			torrent := model.Torrents{
+				Name:         uploadForm.Name,
+				Category:     uploadForm.CategoryId,
+				Sub_Category: uploadForm.SubCategoryId,
+				Status:       1,
+				Hash:         uploadForm.Infohash,
```
majestrate: this should be obtained from the parsed torrent instead

PR #146, opened 2017-05-07
Automate deployment with Ansible (WIP)

tomleb:
# WIP

Don't merge yet.

## Planned
- Installation of docker
- Installation of postgres
- Monitoring system (?) ala Grafana/prometheus
- Alert system (?) Could put nagios on there and alert when something bad happens.
- Docker swarm + provide mean for other to be part of the docker swarm (?) Would help distribute the cost and distribute the load on the servers

I thought maybe we could have one master database with RW access and couple of replicas with RO access, I'll need to do more research.


tomleb:
After discussion, I won't work on docker swarm / replication yet.

thetooth:
I would just provide a prometheus exporter for metrics and leave it up to a external prometheus server to do the alerting. Ideally the codebase itself should be instrumented rather than using an exporter.

As for docker swarm, I don't see how this would be a benefit to most users, HA is very application specific.

tomleb:
Do you mean that you'd not use Nagios but just use prometheus' metrics for alerting ? Docker swarm would be useful to distribute the load on the servers. But yeah, we're not planning to use it yet.

tomleb:
TODOs for backup:
- Make .torrents file out of the dump + signature  (Which do I put for announcement URL with mktorrent ?)
- Seed the files
- Clean up dumps older than X days
- Add auto gen of GPG keys
- Add documentation for developers that want to deploy
- Put public GPG key on the website

tomleb:
Ansible script will fail for rtorrent.service, need to fix this tomorrow, now sleep time.

tomleb:
Need to change some settings in pg_hba.conf to be able to use pgpool.
The application also need to connect to -h localhost -p 9998 instead of -p 5432 to use pgpool.
Parallel queries are not enabled yet but memchache are.

tomleb:
Doing some experimentation with pgpool's memcache. Still need to figure out how big we want our cache to be.
```
nyaapantsu=# select * from torrents LIMIT 1000;
Time: 2137.521 ms
nyaapantsu=# select * from torrents LIMIT 1000;
Time: 5.133 ms
```

```
nyaapantsu=# select * from torrents where torrent_name LIKE '%HorribleSubs%';
Time: 10630.979 ms
nyaapantsu=# select * from torrents where torrent_name LIKE '%HorribleSubs%';
Time: 70.363 ms
```
```
nyaapantsu=# select * from torrents LIMIT 350000;
Time: 6677.070 ms
nyaapantsu=# select * from torrents LIMIT 350000;
Time: 732.065 ms
```

yiiTT:
Nice

tomleb:
Can we reopen it ? I wasn't done and I'm keeping track of what I'm doing here.

PR #150, opened 2017-05-07
don't use type=number

majestrate:


majestrate:
oopps this also includes stuff for issue #151

sfan5:
this fix is incorrect: torrents with unknown filesize should not be ignored completely
the real cause if that those columns are `NULL` which causes gorm to return 0 but it messes with the sorting anyway

the easiest (but not proper) fix would be to ignore size=`NULL` entires only when sorting by filesize

PR #154, opened 2017-05-07
Json search results

majestrate:
allow serach results to be in json if specified by query parameter `t=json`


sfan5:
We already have an api that returns JSON (http://localhost:9999/api/1), that should be improved to allow search queries instead.

Also this includes two unrelated commits

PR #156, opened 2017-05-07
Upload fixes

azhao12345:
we should validate the captcha before processing the upload to protect better from potential attacks

### DIFF
--- a/router/upload.go
+++ b/router/upload.go
@@ -10,9 +10,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
-	"regexp"
 
 	"github.com/ewhal/nyaa/config"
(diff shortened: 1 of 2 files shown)


router/upload.go
```diff
@@ -109,7 +113,6 @@ func (f *UploadForm) ExtractInfo(r *http.Request) error {
 		return ErrInvalidTorrentCategory
 	}
 
-
```
sfan5: the 2 blank lines here are supposed to split the function into logical units...

azhao12345: I just ran go fmt and it stripped the extra whitespace.  Not sure if trying to keep non-standard whitespace is sustainable at all.

majestrate: if `go fmt` removed it it should stay removed

sfan5: go sucks ass confirmed

PR #168, opened 2017-05-07
Unfuck shit

sfan5:
can we stop blindly merging stuff?

this also means #153 needs to be redone


aerojun:
What happened with the fucked merges?

PR #175, opened 2017-05-08
implement comments

azhao12345:
initial implementation


azhao12345:
should be done enough now probably

PR #189, opened 2017-05-08
changes

sfan5:


andrewzhaosc:
If we were going to change the schema to put the old comments into a new table, why didn't we just merge the old comments and new comments into one table in the migration?

andrewzhaosc:
or since we don't have any "new comments" that are real yet, can't we just make one comment model and point it to the migrated table?

It seems poor to have to seperate tables that are basically holding the same stuff

ewhal:
This is still loading the old comments but it doesn't look for old comments on new torrents

sfan5:
@andrewzhaosc
 The point of not merging the comments table is that I don't want to lots of shit (all nyaa accounts) into the `users` table.

andrewzhaosc:
@sfan5  ok that's fair.  I guess the new comments have the commentser userid, while the old ones just have a username slapped in

sfan5:
Indeed, `comments_old` and `user_uploads_old` are just read-only views of data that can't be reasonably merged with the new stuff.

PR #192, opened 2017-05-08
Death to reallocations!

bakape:


bakape:
Reduces backing array reallocation, when appending slices.

PR #203, opened 2017-05-09
Fixed searching in sqlite db

DEC05EBA:
sqlite doesn't support ILIKE -> changed to UPPER(columnName) LIKE
'%STRING%'


util/search/search.go
```diff
@@ -113,7 +113,7 @@ func SearchByQuery(r *http.Request, pagenum int) (SearchParam, []model.Torrents,
 			continue
 		}
 		// TODO: make this faster ?
-		conditions = append(conditions, "torrent_name ILIKE ?")
+		conditions = append(conditions, "UPPER(torrent_name) LIKE ?")
```
Atvaark: You'd also need to UPPER the parameter and hope there's an index for *UPPER(torrent_name)* or there'll be a full table scan.

Also some dbs (but not postgres) can be configured to compare case insensitively.

Issue #199, opened 2017-05-09
Night mode does not theme the member menu


ripdog:
<img width="218" alt="nightmode" src="https://cloud.githubusercontent.com/assets/174944/25835213/03026ef2-34d1-11e7-8a88-eb87ef1b7f63.PNG">

Firefox Nightly 55, Win10

yiiTT:
Night theme moves top search bar due to span.search_text's contents "Search" **and** also keep the switcher in the same location to prevent further issues with alignment and increase usability.

![](http://i.imgur.com/C6zS277.png)
![](http://i.imgur.com/e5rFSHK.png)

ripdog:
Fixed with #205

PR #205, opened 2017-05-09
CSS rework - additive night mode CSS, and use CSS background image for night mode toggle

ripdog:
Currently, style-night.css is a copypaste of style.css, but with different colours. This is very bad design, as any layout changes must be replicated on both files. Already this has caused a difference in layout on the dark theme. My change removes all layout instructions from the dark mode css, and adds it on top of the existing css. This guarantees that no layout changes are made by dark mode. 

Also fixed the dropdowns in night mode.

Also displaying the night-mode toggle button with a css background instead of an img. This nets 2 lines less JS.

### DIFF
--- a/public/css/style-night.css
+++ b/public/css/style-night.css
@@ -1,8 +1,4 @@
-/* Night mode switcher  */
-.nightswitch > img {
-    width: 24px;
-}
-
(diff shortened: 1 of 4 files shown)


ripdog:
Fixed #199 for realz this time

PR #217, opened 2017-05-09
Various changes

### LINKED ISSUE #201: Gravatar hash is not being generated
See https://nyaa.pantsu.cat/user/3/yiiT and you will find the avatar src to be "https://www.gravatar.com/avatar/?s=130"

sfan5:
should definitely fix #201

### DIFF
--- a/config/upload.go
+++ b/config/upload.go
@@ -1,8 +1,14 @@
 package config
 
 const (
-	// TorrentFileStorage = "/var/tmp/torrent_outgoing"
+	// TorrentFileStorage = "/var/www/wherever/you/want"
+	// TorrentStorageLink = "https://your.site/somewhere/%s.torrent"
 	TorrentFileStorage = ""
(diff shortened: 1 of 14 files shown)


config/upload.go
```diff
@@ -1,8 +1,14 @@
...
 const (
-	// TorrentFileStorage = "/var/tmp/torrent_outgoing"
+	// TorrentFileStorage = "/var/www/wherever/you/want"
+	// TorrentStorageLink = "https://your.site/somewhere/%s.torrent"
 	TorrentFileStorage = ""
+	TorrentStorageLink = ""
+
+	// TODO: deprecate this and move all files to the same server
+	TorrentCacheLink = "http://anicache.com/torrent/%s.torrent"
+
```
alucard0134: whitespace AAA

sfan5: at least it was tabs and not spaces

PR #222, opened 2017-05-09
Search caching

bakape:
Not sure if we still need caching of `[]model.Torrents`. There is caching on our deployment through PGpool, but it will likely not apply to any mirrors. Waiting on @ewhal for decision.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -1,0 +1,125 @@
+package cache
+
+import (
+	"container/list"
+	"sync"
+	"time"
+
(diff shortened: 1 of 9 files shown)


majestrate:
mor cashing is good

bakape:
Done. This fill cache search, RSS and home page request results in memory for 1 minute without querying the DB. 
Also prevents flooding, when several clients request the same resource (like the default home page) in a short span of time. If there is no fresh cache record, only the first client will make a DB query and the rest will block, until that finishes, and then get a copy.

PR #236, opened 2017-05-09
Get custom 404 page working for all handlers

SpamNeko:
Currently wraps functions intended for HandlerFunc and passes off to NotFoundHandler for all 404s, but NotFoundHandler doesn't play nice with gzip.

I'll probably try to make it wrap actual Handlers instead.


SpamNeko:
This now seems to be working, but I'll keep it WIP until I can test it a bit more later.

SpamNeko:
This seems to be good to go, fixes [https://github.com/ewhal/nyaa/issues/135](https://github.com/ewhal/nyaa/issues/135) and [https://github.com/ewhal/nyaa/issues/227](https://github.com/ewhal/nyaa/issues/227).

PR #243, opened 2017-05-10
Workaround large torrent IDs on viewTorrentHandler giving blank pages

### LINKED ISSUE #227: Trying to access torrents with IDs over 32-bit Integer limit yields blank page
[On 2,147,483,647](https://nyaa.pantsu.cat/view/2147483647), it fails (although there's no Page not Found, but that's another problem).
[On 2,147,483,648](https://nyaa.pantsu.cat/view/2147483648), it succeds with a blank page.

ElegantMonkey:
Works, but I wonder why the DB won't trigger RecordNotFound. With #236, should fix #227.


alucard0134:
ill wait until #236 is actually done

PR #262, opened 2017-05-10
Revert "New view template design"

azhao12345:
Reverts ewhal/nyaa#256


aerojun:
Why was it reverted so i can fix it.

ewhal:
@aerojun I don't recall but you can fix it and make a pr again.

PR #291, opened 2017-05-10
Add template and db tests

Atvaark:
- The db test will run an sqlite3 database in in-memory
  mode and run the automigrations.
- The template test will check if all the templates
  that are stored in package variables (template.Must)
  compile and if ReloadTemplates works without panicking.

### DIFF
--- a/db/gorm_test.go
+++ b/db/gorm_test.go
@@ -1,0 +1,24 @@
+package db
+
+import (
+	"testing"
+
+	"github.com/ewhal/nyaa/config"
+)
+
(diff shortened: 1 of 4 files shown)


alucard0134:
you should switch this to postgres since sqlite is getting ripped soon, but ill allow it

Atvaark:
I haven't if Travis CI supports setting up databases for integration tests, but it looks like they do.
https://docs.travis-ci.com/user/database-setup/

Once sqlite support is dropped the db test should be easily ported to postgres.

PR #296, opened 2017-05-10
Bringing back megumin

S0X-LT:
dunno if this works, but compared it to style.css and it should work.

### DIFF
--- a/public/css/style-night.css
+++ b/public/css/style-night.css
@@ -66,7 +66,7 @@
 */
 img#mascot {
 	display:block;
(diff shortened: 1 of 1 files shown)


alucard0134:
test if it works first pls

kipukun:
you piece of shit don't use css backgrounds it's impossible to work with

kipukun:
And Megumeme is already in the latest HEAD. This can be closed.
fucking

S0X-LT:
alright then

PR #317, opened 2017-05-10
Japanese translation

Redworthy:


ElegantMonkey:
Those `”` quotes should be normal `"` quotes.

Redworthy:
Fixed.
Not sure why the translator changed the quotations...

S0X-LT:
about time

PR #325, opened 2017-05-10
Show date created for comments. (don't merge yet!!!)

TUSF:
Hopefully I didn't break something. I don't normally use Bootstrap.

### DIFF
--- a/templates/view.html
+++ b/templates/view.html
@@ -67,12 +67,17 @@
                 <h4>{{T "comments"}}</h4>
                 {{ range $index, $element := .Comments }}
                 <div class="row comment-row" id="comment_{{$index}}">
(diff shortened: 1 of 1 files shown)


ayame-git:
borked i think

ayame-git:
O-o-ok

aerojun:
I don't know if it's because i'm using the sqlite db but i get the following error:

`ERRO[0001] ViewHandler(): template: view.html:97:36: executing "content" at <.CreatedAt.Year>: can't evaluate field CreatedAt in type model.CommentJSON`

TUSF:
@aerojun Crap; I used the wrong model, and thus the wrong field name. Hold up.

PR #346, opened 2017-05-11
Fix ja-jp.all.json

MMP0:


TKR2:
Thank you for a more accurate Japanese translation.

PR #349, opened 2017-05-11
Redoing displayed Comment format.

TUSF:
It was showing it as 2017/May/11, which is just ugly. Also changed it up so month and day number are 0-padded, so instead of just "2017-5-11" it'll show "2017-05-11".


TUSF:
Ok, **NOW** it works. Didn't realize text/template didn't allow spaces before and after comments (`{{/* */}}` vs (`{{ /* */ }}`)

PR #352, opened 2017-05-11
Add marquee for FAQ.html

MMP0:
&lt;marquee&gt; is obsolete


sfan5:
38 lines of CSS instead of a single HTML tag that works in all browsers anyway?
nice meme

PR #370, opened 2017-05-11
Change color for bg-danger and fix night theme flickering issue

tomleb:
Overrides bootstrap's default bg-danger color because it looks ugly, especially using the night theme.
Here's the old version: [https://my.mixtape.moe/hktvzt.png](https://my.mixtape.moe/hktvzt.png)
Here's the new version: [night](https://my.mixtape.moe/wsuhvc.png) and [normal](https://my.mixtape.moe/zyrjkv.png)

We can change the colors if any of you still find it ugly, as long as we don't make it default.

### DIFF
--- a/public/css/style-night.css
+++ b/public/css/style-night.css
@@ -145,7 +145,7 @@ a:hover {
     color: #fff;
 }
 
-.bg-danger {
+.text-error {
   background: #29363d;
   color: #cf9fff;
 }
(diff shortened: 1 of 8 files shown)


akuma06:
Why don't you use instead .text-info of modifying .bg-danger ?

tomleb:
I'll change .bg-danger to something else and make a template to avoid duplication of the inline script as well before merging this.

PR #411, opened 2017-05-13
Rethinking the design a bit (WIP)

tomleb:
Mostly a PR to deal with this [https://github.com/ewhal/nyaa/issues/383](https://github.com/ewhal/nyaa/issues/383)

Looks like this for the moment: [https://my.mixtape.moe/skdwdw.png](https://my.mixtape.moe/skdwdw.png)

I'm pretty bad at html/css, had to use inline flex box at ~2 places for vertical alignment.

Done:
- Permanently visible advanced searched
- Only show advanced search for the index
- Moved night theme to the right

Don't merge yet, still some stuff to do:
- Fully center advanced search
- Get rid of weird line in the upload page
- Sign up and sign on seems to have different vertical spacing
- A bit of spacing between mascot and nav bar + reduce mascot's size
- From Chagall: FIX THIS [https://i.imgur.com/x03jAmd.png](https://i.imgur.com/x03jAmd.png)
- From ReimuHakur: Space between caret and member label


tomleb:
Now that the advanced research is gone, I'll be implementing the design showed [here](https://github.com/ewhal/nyaa/issues/464) instead, which looks a lot cleaner imo.
The new page will look like this:
New design

nakedloli:
You should implement his too #484

tomleb:
Here's what I have now.
Light: 
light
dark

- Should I keep pagination of 10 or 15?
- Should we change the labels (dark/light) to our previous icons or keep this and translate it to other languages ? If we decide to go with the icons, we need to make them black (or at least darker than what they are now)
- I'm not too convinced about how the mascot looks when the width is small. Then again our website isn't too much good looking for mobiles yet. (see below)
mascot 
- Keep or remove the small space between the navbar and the black container?

Any thoughts ?

autismbux:
It's looking quite good!

For the pagination, I was thinking maybe consider letting it go from 1 to 5, then increase them in increments of 5 after that? So it would be [1, 2, 3, 4, 5, 10, 15, 20]. This would make it easier for both short and medium range page jumping. Is that easy to do?

kipukun:
To answer your questions respectively:
- Keep pagination at 10. 
- Keep the icons. Should be easy to fix with some CSS applied to them, and we wouldn't need to TR.
- http://getbootstrap.com/css/#responsive-utilities
- Increasing the space would make it more separate from the navbar, so I say keep it that way.

xhocquet:
@tomleb I'm quite good at CSS, if you need help fleshing some of this out in a proper CSS framework, let me know!

PR #415, opened 2017-05-13
Move 'advanced' search into table headers

SpamNeko:
See comment at bottom before merging.

The so-called 'advanced' search is actually just a duplication of elements elsewhere on the page, except pagination which isn't even really a search option (from user's perspective). See [diagram](http://puu.sh/vOklD/9d166d6eaf.png).

I'm moving the sorting options into the header of the table, so it can be sorted the same way as every other not-terrible HTML table ever. (i.e., click header to sort by that, click again to reverse order)

To Do:
- DELET advanced search
- Handle blank query string properly
- Up/Down arrows to show what we are sorting by? (see also: the other nyaa)
- Determine how this relates to torrent list on user profile page.
- Rebase

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -202,6 +202,10 @@ div.container div.blockBody:nth-of-type(2) table tr:first-of-type th:last-of-typ
 {
 	width: 12rem;
 }
+#mainmenu .navbar-form select.form-control#max
+{
+  width: 8rem;
+}
 .special-img 
 {
     position: relative;
--- a/router/templateFunctions.go
+++ b/router/templateFunctions.go
@@ -36,6 +36,24 @@ var FuncMap = template.FuncMap{
 		}
 		return "error"
 	},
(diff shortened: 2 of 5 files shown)


tomleb:
Seem to be duplicate of https://github.com/ewhal/nyaa/pull/411 kinda

SpamNeko:
Yeah, I noticed there was working going on on the design, so I'm only going to move the sorting into the headers and not touch any other elements of design for now.

SpamNeko:
This now works, it removes advanced search and instead lets you click table headers to sort.

There are no arrows on the table headers (yet), so don't merge if you think users are too stupid to spot that the table headers are links. Either way, I'll keep working on it.

Known issues: Sorting links don't do anything on home page due to confusion between home/search routes. No arrows. Default sort is by ID which is not even an option anymore, should probably be by date.

PR #428, opened 2017-05-13
Fix registration view tab order

Atvaark:
Since the captcha uses the default tab order it was getting skipped
by the registration form.

- Add autofocus to the registration and upload view
- Add id attribute to input elements that have a label


sfan5:
doesn't the `for` attribute of a `<label>` refer to the `name` and not the `id`?

Atvaark:
https://developer.mozilla.org/en/docs/Web/HTML/Element/label

> **for** The ID of a labelable form-related element in the same document as the label element.

PR #440, opened 2017-05-13
Add filesize fetcher mode (need SOMEONE for testing)

ElegantMonkey:
Uses a torrent client to fetch the filesize of the torrents with `filesize IS NULL` or `filesize = 0` on the DB.

Seems to work, but needs to be tested with a more recent DB (especially with the HorribleSubs scraped one)

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ewhal/nyaa/network"
 	"github.com/ewhal/nyaa/router"
 	"github.com/ewhal/nyaa/service/scraper"
+	"github.com/ewhal/nyaa/service/torrent/filesizeFetcher"
 	"github.com/ewhal/nyaa/util/log"
 	"github.com/ewhal/nyaa/util/search"
 	"github.com/ewhal/nyaa/util/signals"
@@ -94,6 +95,18 @@ func RunScraper(conf *config.Config) {
 	scraper.Wait()
 }
 
+// RunFilesizeFetcher runs the database filesize fetcher main loop
+func RunFilesizeFetcher(conf *config.Config) {
+	fetcher, err := filesizeFetcher.New(&conf.FilesizeFetcher)
+	if err != nil {
+		log.Fatalf("failed to start fetcher, %s", err)
+		return
+	}
+
+	signals.RegisterCloser(fetcher)
+	fetcher.Run()
+}
+
 func main() {
 	conf := config.New()
 	processFlags := conf.BindFlags()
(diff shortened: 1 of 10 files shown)


tomleb:
Shouldn't that be a different application instead ?

ElegantMonkey:
@tomleb: It is, kinda. You run it alongside the nyaa webapp, just like the scrapper, with `-mode filesize_fetcher`.

majestrate:
I like this a lot

main.go
```diff
@@ -142,6 +156,8 @@ func main() {
 			RunScraper(conf)
 		} else if *mode == "webapp" {
 			RunServer(conf)
+		} else if *mode == "filesize_fetcher" {
```
majestrate: nitpick: should be `metainfo_fetcher` instead as it fetches more than just the file sizes

ElegantMonkey: Yeah, but for now it only updates the filesize, so I think it's more appropriate calling it like this. But I can rename it if you want to.

akuma06:
Just resolved your conflicts

ElegantMonkey:
I would like to test on a recent DB really, only tested it with 1-2 torrents, as there are few recent torrents with no filesizes on the old one. Otherwise, I guess not.

akuma06:
If you have tested it on 2 torrents, then it should be ok. I don't see why it would broke.

ElegantMonkey:
Ok then, I hope it's fine.

ewhal:
This is neat if we could also do something like this but for filelists then that would be really great!

ElegantMonkey:
@ewhal, you mean the files inside the torrents? Probably could be done, but I don't know how we should store it (I don't think there's a field for it right now).

ewhal:
We can always add a column to the db.
But yeah for the files in the torrent.

ElegantMonkey:
@ewhal, I'll take a look, but could you test this with your DB? I've changed a few torrents manually to `filesize = 0` for testing, but for some of them I get 0 bytes as the length...

ewhal:
sure will do and thank you for your help!

ElegantMonkey:
Alright, now it reads the FileInfo from the metadata too. Probably should rename it, like @majestrate said, but I think it's good for now...

ewhal:
Just tested it and it works

ewhal:
@ElegantMonkey I didn't see that you had added filelist support and I just retested it and it doesn't work.
It successfully gets missing size information though just not the list of files per torrent.

PR #447, opened 2017-05-13
Fix i18n file error

Atvaark:
- Fixed an error in the spanish and chinese translation file
- Refactored reading the translation files
  Moved the code out of main.go and instead of ignoring
  errors in the non-default language files
  the error will now be returned and the remaining files will
  not be loaded.
  **I could revert this to the previous logic (load all files) and then return an aggregated error** Suggestions?
- Added a unit test to check if all translation files are valid
- Added an i18n config to specify the translations dir and the
  default language

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -21,7 +21,7 @@ type Config struct {
 	DBType string `json:"db_type"`
 	// DBParams will be directly passed to Gorm, and its internal
 	// structure depends on the dialect for each db type
-	DBParams string `json:"db_params"`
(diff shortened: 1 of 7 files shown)


ElegantMonkey:
> Added a unit test to check if all translation files are valid

God bless.

akuma06:
Still Wip?

Atvaark:
You can merge this PR.

I'll take a look at the TODO some time later (lots of hardcode "en-us" literals)

PR #452, opened 2017-05-13
Adding our tracker to FAQ

S0X-LT:
Can someone please test this? i mean i just edited simple HTML, but i can't test it myself and don't want to break stuff.

Also, we'll need to update the translations.


akuma06:
Would that not be better to still keep some other trackers as a surplus list? So if the tracker address change, the torrents will still work even if people don't edit the tracker url in their torrent file.

S0X-LT:
kk lemme edit real quick

S0X-LT:
Alright, done.

PR #477, opened 2017-05-14
MetainfoFetcher and file lists

ElegantMonkey:
A few changes from #440:
- Renamed `FilesizeFetcher` to `MetainfoFetcher`.
- Updated logic so that it gets the metainfo of both torrents missing file sizes or missing file lists.
- Show file lists on the View Torrent page.
- Read file list from uploaded .torrent files.
- Make failed fetches be able to be fetched again, with a exponential cooldown.

~~I couldn't test this because my connection is really trash right now~~ Tested it, seems to work fine albeit a bit slow, might need to experiment with Timeout/Queue size. PR'ing for thoughts/suggestions.


ewhal:
Seems to be working but I haven't tested it on any batches yet

service/torrent/metainfoFetcher/metainfoFetcher.go
```diff
@@ -150,12 +152,28 @@ func (fetcher *MetainfoFetcher) gotResult(r Result) {
...
+	if fetcher.failCooldown < 0 {
+		return
+	}
+
+	now := time.Now()
+	for id, failTime := range fetcher.failedOperations {
+		if failTime.Add(time.Duration(fetcher.failCooldown) * time.Second).Before(now) {
+			log.Infof("Torrent TID %d gone through cooldown, removing from failures")
+			// Deleting keys inside a loop seems to be safe.
+			delete(fetcher.failedOperations, id)
```
majestrate: pretty sure this will crash

ElegantMonkey: I was afraid too, but after some Googling it seems to be safe. [https://github.com/golang/go/issues/9926](url)

majestrate: is the map only accessed from 1 goroutine ?

ElegantMonkey: This map is only accessed by functions called from `MetadataFetcher.run()`, which is a single goroutine. The spawned goroutines don't access it, AFAIK. Could use a Mutex just for safety, though.

PR #491, opened 2017-05-14
Updated translation.

Myrmece:


translations/fr-fr.all.json
```diff
@@ -544,6 +556,10 @@
     "translation": "Modérateur"
   },
   {
+    "id":" api_token",
```
sfan5: should be `    "id": "api_token",`

PR #499, opened 2017-05-14
Adds pt-pt and fixes pt-br

CrisBRM:


mitki:
In brazilian portuguese "porquê" is only used as a noun, so the correct is "por que".

CrisBRM:
Ignore that commit then. It sounded odd to me but I am aware of my own limitations when it comes to pt-br.

akuma06:
Just did that =). Thanks for the translation by the way

PR #504, opened 2017-05-14
Sorting arrows on table headers

SpamNeko:
Sorting arrows are in place but do not really fit with the current layout.

[http://puu.sh/vPZ22/422ab1ef6d.png](http://puu.sh/vPZ22/422ab1ef6d.png)

See [https://github.com/ewhal/nyaa/issues/484](https://github.com/ewhal/nyaa/issues/484)

### DIFF
--- a/public/css/style.css
+++ b/public/css/style.css
@@ -136,21 +136,23 @@ tr.torrent-info td.date {
   border-top: 1px solid #ddd;
 }
 
+th a:hover {
+  text-decoration: none;
+}
+
+.sortarrowleft {
+  letter-spacing: -0.6rem;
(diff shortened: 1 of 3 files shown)


akuma06:
If you want to do with separate colomn, PR it =).

SpamNeko:
S/L/D is now multi-column, and much nicer arrows. [Image](http://puu.sh/vQc2l/7e1492bd58.png)

To make the layout not completely break, I also changed the column class on the name column from 8 to 12. The layout might need some fixing up, but there seems to be upcoming work on changing the theme significantly anyway.

PR #506, opened 2017-05-14
Replace 'en-us' literals with the default language

Atvaark:
This consolidates the places where a default language has to be set.

- Removed import of the 'userService' package into the 'languages' util
  package
  This was required to prevent a cyclic import between the two packages.
- Added a 'UserRetriever' interface to read the language setting of users
  inside the 'languages' package


akuma06:
Seems to be good

PR #510, opened 2017-05-14
[TL] Update pt-br

mitki:
Updated~


akuma06:
Waiting travis compile after fix

Issue #451, opened 2017-05-13
Add more RSS fields


misuzu:
Right now there is ```title```, ```link```, ```description```, ```guid```, and ```pubDate```.
nyaa.eu RSS also had ```category``` field.
[EZTV RSS](https://eztv.ag/ezrss.xml) have even more useful fields: ```torrent:contentLength```, ```torrent:infoHash```, ```torrent:magnetURI```, ```torrent:seeds```, ```torrent:peers```, ```torrent:verified```, ```torrent:fileName```.
Direct link to torrent file would also be great.

ipatrol:
We should make sure to be compliant with the BEP standard for torrent RSS feeds: http://www.bittorrent.org/beps/bep_0036.html

yiiTT:
#569 "Implement EZTV RSS Spec"

PR #542, opened 2017-05-16
Added torrent rss fields

Akaahn:
As suggested by #451.
Pulled  gorilla/feeds into nyaa/feeds and modified to allow for `torrent` tags.
It now supplies
`torrent:seeders`
`torrent:leechers`
`torrent:hash`
`torrent:completed`
`torrent:lastScrape`

### DIFF
--- a/feeds/.travis.yml
+++ b/feeds/.travis.yml
@@ -1,0 +1,14 @@
+language: go
+sudo: false
+go:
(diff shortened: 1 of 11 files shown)


yiiTT:
Hmm any reason this was closed? We will be implementing the EZTV RSS Spec #569

PR #547, opened 2017-05-16
Romanian translation

fuzzyballz:


fuzzyballz:
Cum = How and not actual cum btw.

PR #592, opened 2017-05-17
Fix #570

### LINKED ISSUE #570: Sorting by seeds when all relevant returned have 0 seeds returns empty list
Currently if a search term yields valid results (but none have seeds) then sorting by seeders returns an empty page.  Feel like it'd be less confusing to just maybe return the same page or maybe have some other indicator that you ARE sorting by seeds and the maximum seeds is 0.  Not sure if this is a bug or a design choice though.  Either way I feel like it should be changed.

akuma06:


akuma06:
Yes it does work and yes I have seen your comment. HOWEVER, when you say "break", I don't understand what it means, so for me that was something absurd for a query breaking because of no order parameter

akuma06:
The fact that null values comes first with a Desc order have been a normal behavior for postgres. I just forgot about it and didn't add LAST or FIRST to tweak it.

PR #608, opened 2017-05-18
Update Japanese translations

MMP0:


translations/ja-jp.all.json
```diff
@@ -480,6 +480,10 @@
     "translation": "二次元 - 漫画"
   },
   {
+    "id": "art_games",
+    "translation": "二次元 - 画像"
```
sfan5: duplicate, see 8 lines above

MMP0: Thanks

PR #611, opened 2017-05-18
Update FAQ.html

yiiTT:
Someone overwrote these changes during a merge


sfan5:
This is why we should be using rebased commits instead of merges.

goonyaa:
@yiiTT looks like https://github.com/NyaaPantsu/nyaa/pull/508 was merged straight into master, while https://github.com/NyaaPantsu/nyaa/pull/486 was in dev branch. Strange things.

PR #615, opened 2017-05-18
Adding more options for upload disabling

Akatsuki-Ousawa:
Added a few more options for Upload disabling.

### DIFF
--- a/router/apiHandler.go
+++ b/router/apiHandler.go
@@ -96,13 +96,19 @@ func ApiViewHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func ApiUploadHandler(w http.ResponseWriter, r *http.Request) {
-	if config.UploadsDisabled {
+	user := model.User{}
+	if config.UploadsDisabled && config.AdminsAreStillAllowedTo && user.Status != 2 && config.TrustedUsersAreStillAllowedTo && user.Status != 1 {
+		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
+		return
+	} else if config.UploadsDisabled && !config.AdminsAreStillAllowedTo && user.Status == 2 {
+		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
+		return
+	} else if config.UploadsDisabled && !config.TrustedUsersAreStillAllowedTo && user.Status == 1 {
 		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
 		return
 	}
 
 	token := r.Header.Get("Authorization")
-	user := model.User{}
 	db.ORM.Where("api_token = ?", token).First(&user) //i don't like this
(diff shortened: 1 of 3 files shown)


router/apiHandler.go
```diff
@@ -173,15 +179,21 @@ func ApiUploadHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func ApiUpdateHandler(w http.ResponseWriter, r *http.Request) {
-	if config.UploadsDisabled {
+	user := model.User{}
```
sfan5: This is an empty struct, you need to actually read the user from the db

router/apiHandler.go
```diff
@@ -96,13 +96,19 @@ func ApiViewHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func ApiUploadHandler(w http.ResponseWriter, r *http.Request) {
-	if config.UploadsDisabled {
+	user := model.User{}
```
sfan5: same applies here

router/apiHandler.go
Akatsuki-Ousawa: But where is it read in the old file? I mean I just moved it up a few lines and from my understanding nothing redecleared it to fill it

sfan5: https://github.com/NyaaPantsu/nyaa/pull/615/files#diff-021a8ffd4664e7215a6a15d63ed60234R197

Akatsuki-Ousawa: changed it a bit pls take a look

router/apiHandler.go
```diff
@@ -96,14 +96,20 @@ func ApiViewHandler(w http.ResponseWriter, r *http.Request) {
...
+	if config.UploadsDisabled && config.AdminsAreStillAllowedTo && user.Status != 2 && config.TrustedUsersAreStillAllowedTo && user.Status != 1 {
+		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
+		return
+	} else if config.UploadsDisabled && !config.AdminsAreStillAllowedTo && user.Status == 2 {
+		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
+		return
+	} else if config.UploadsDisabled && !config.TrustedUsersAreStillAllowedTo && user.Status == 1 {
+		http.Error(w, "Error uploads are disabled", http.StatusBadRequest)
+		return
+	}
```
sfan5: Hm, the config.UploadsDisabled should probably be moved to a seperate if-clause for better readability

Akatsuki-Ousawa: So more like an outer if clause for config.UploadsDisabled and an inner for the rest?

sfan5: Yep

Akatsuki-Ousawa: changed it

PR #620, opened 2017-05-18
Possible fix for torrents order on user profile page

### LINKED ISSUE #396: Torrent list on user page is ordered by date ascending
should be descending really

goonyaa:
Changes query from:
```sql
SELECT * FROM "torrents"  WHERE "torrents"."deleted_at" IS NULL AND (("uploader" IN ('1'))) ORDER BY "torrents"."torrent_id" ASC
```
to:
```sql
SELECT * FROM "torrents"  WHERE "torrents"."deleted_at" IS NULL AND (("uploader" IN ('1'))) ORDER BY "torrents"."torrent_id" DESC
```

Should close #396


yiiTT:
Confirmed Fix

sfan5:
Interesting fix right there, probably some internal GORM behaviour...

PR #621, opened 2017-05-19
Make comments extra column + search bar doesn't freak out anymore

### LINKED ISSUE #618: Long torrent names get messed up with comment badge

kipukun:
Fixes #618. Torrent name column is small due to there being no s/l/d data. It'll get bigger when those fill.


kipukun:
Forgot image!
capture

PR #627, opened 2017-05-19
Add tooltips to category images

moose65468:
Adds tooltips to the category images which also work across translations
I did this by adding a couple functions which use the category and sub-category num's

pop-up

### DIFF
--- a/router/template_functions.go
+++ b/router/template_functions.go
@@ -163,4 +163,51 @@ var FuncMap = template.FuncMap{
 		// because time.* isn't available in templates...
 		return t.Format(time.RFC3339)
 	},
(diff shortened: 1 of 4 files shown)


router/template_functions.go
```diff
@@ -163,4 +163,51 @@ var FuncMap = template.FuncMap{
...
+            case "6_15":    e = "pictures_graphics"
+            case "6_16":    e = "pictures_photos"
+            case "1_":      e = "software"
+            case "1_1":     e = "software_applications"
+            case "1_2":     e = "software_games"
+        }
+        return e
+    },
+}
 }
```
MMP0: syntax error, "}" on line 213

moose65468: oops

PR #632, opened 2017-05-19
Fix a tiny typo on documentation

hikaru-light:
Thank you for your watching.
Maybe there is a tiny mistake.
Please check this.


tomleb:
I think it should be 'This is purely an educational proof of concept.'

hikaru-light:
Oh, thank you for your advice.
If it is so, I will close this PR.
Sorry...

hikaru-light:
I considered it.
That seems reasonable to me.

PR #652, opened 2017-05-20
Move translation func to template variables

ElegantMonkey:
Apparently [template.ExecuteTemplate doesn't lock the templates' FuncMap](golang.org/src/text/template/exec.go?s=5133:5197#L181), meaning that the FuncMap can change during the template execution. As the current translation method is changing the translation func on the FuncMap, this means the language can change too.

Hopefully this will fix the bug with mixed translations on the site.

### DIFF
--- a/router/database_dump_handler.go
+++ b/router/database_dump_handler.go
@@ -49,8 +49,8 @@ func DatabaseDumpHandler(w http.ResponseWriter, r *http.Request) {
 
 	// TODO Remove ?
 	navigationTorrents := Navigation{0, 0, 0, "search_page"}
-	languages.SetTranslationFromRequest(databaseDumpTemplate, r)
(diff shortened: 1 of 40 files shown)


yiiTT:
Captcha is broken with this pull
```
template: _captcha.html:5:32:
executing "captcha" at <$.T>:
can't evaluate field T in type
router.UploadForm
```

ElegantMonkey:
@yiiTT, sorry, should be fixed with #661. Is it the only broken thing you've noticed?

ElegantMonkey:
Or... nevermind, it's still broken on the other pages. I'll look at it.

yiiTT:
@ElegantMonkey looks like there is still some issues everywhere else _captcha is used for Commenting and Registering,

ElegantMonkey:
@yiiTT Yeah, I've seen it - I'll add a template function to bundle the CaptchaID and the Tfunc on a struct, which _captcha will use. Aside from this, anything else?

yiiTT:
@ElegantMonkey nope that should be it, captcha is used for registering, commenting, and uploading.

PR #653, opened 2017-05-20
Torrent weblink

akuma06:


router/upload.go
```diff
@@ -116,6 +123,12 @@ func (f *UploadForm) ExtractInfo(r *http.Request) error {
 		return ErrInvalidTorrentCategory
 	}
 
+	// WebsiteLink
+	urlRegexp, _ := regexp.Compile(`^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$`)
```
sfan5: what about IRC links?

akuma06: There was?

sfan5: Yes this was supported on old nyaa

akuma06:
fixed

PR #654, opened 2017-05-20
Update for css formatting, ru-ru translation and small changes to templates

goonyaa:


public/css/style.css
```diff
@@ -436,15 +473,15 @@ select#bottom_language_selector {
 
 .comment-date {
 	font-size: smaller;
-	width: auto; /* Undo bootstrap's fixed width */
+	/* Undo bootstrap's fixed width */
```
sfan5: Why move the comments? :confused:

goonyaa: Just don't like end of line comments. Can revert this part.

akuma06:
Resolved conflicts.

public/css/style.css
akuma06: I prefer end of line comments with // and new line comments with /* */. So I'm more for his way of doing it

PR #662, opened 2017-05-21
Fix captcha again

ElegantMonkey:
Now works on View Torrent and Registration too.


akuma06:
Look if it works with reporting also.

ElegantMonkey:
@akuma06, yup, it works.

PR #665, opened 2017-05-21
Group common template variables together.

ElegantMonkey:
Creates a new struct, `CommonTemplateVariables`, that groups most of the fields used by (almost) all the template variables.


yiiTT:
looks like it's all working, now just gota watch out for those auto merges.

PR #668, opened 2017-05-21
Fix navbar bugs

MMP0:


akuma06:
It doesn't work with other languages

public/css/style.css
```diff
@@ -269,14 +267,6 @@ div.container div.blockBody:nth-of-type(2) table tr:first-of-type th:last-of-typ
 	display: none;
 }
 
-#mainmenu .navbar-form select.form-control {
-	width: 12rem;
-}
-
```
akuma06: Bug is here, you remove the max width of select elements in the search form

public/css/style.css
```diff
@@ -269,14 +267,6 @@ div.container div.blockBody:nth-of-type(2) table tr:first-of-type th:last-of-typ
...
-#mainmenu .navbar-form select.form-control#max {
-	width: 6rem;
-}
-
 .special-img {
```
akuma06: Same here

MMP0:
Should be fixed...

PR #685, opened 2017-05-22
Verify torrent category on upload

### LINKED ISSUE #682: Torrents with invalid categories.
Example: https://nyaa.pantsu.cat/view/924512

There's no checking if the torrent category on the form actually exists, it checks only if it is in the format `x_y`, where `x` and `y` are both numbers. You can reproduce this locally by editing the `value` field of the selected `option` on the upload page, and trying to upload.

ElegantMonkey:
Should fix #682

### DIFF
--- a/router/template_functions.go
+++ b/router/template_functions.go
@@ -162,7 +162,7 @@ var FuncMap = template.FuncMap{
 		// because time.* isn't available in templates...
 		return t.Format(time.RFC3339)
 	},
-    "CategoryID": func(category string, sub_category string) string {
+    "CategoryName": func(category string, sub_category string) string {
 		s := category + "_" + sub_category
 
 		if category, ok := categories.GetCategories()[s]; ok {
--- a/router/template_functions.go
+++ b/router/template_functions.go
@@ -10,6 +10,7 @@ import (
 	"github.com/NyaaPantsu/nyaa/config"
 	"github.com/NyaaPantsu/nyaa/service/user/permission"
 	"github.com/NyaaPantsu/nyaa/util"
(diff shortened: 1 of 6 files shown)


router/template_functions.go
```diff
@@ -161,51 +162,14 @@ var FuncMap = template.FuncMap{
...
-            case "5_11":    e = "live_action_raw"
-            case "6_":      e = "pictures"
-            case "6_15":    e = "pictures_graphics"
-            case "6_16":    e = "pictures_photos"
-            case "1_":      e = "software"
-            case "1_1":     e = "software_applications"
-            case "1_2":     e = "software_games"
-        }
-        return e
+    "CategoryID": func(category string, sub_category string) string {
```
sfan5: should probably be `CategoryName`

sfan5:
Also you're missing the case where people could try to post in a parent category (without a subcat selected)

akuma06:
@sfan5 Shouldn't it work like that? If someone post in Anime Category, it should check "3_". It exists so it uploads

router/template_functions.go
akuma06: I agree, it would be more fitted to the function

ElegantMonkey:
@sfan5 @akuma06 The upload handler splits the category on the _ sign, and try to convert each string part to an int. The current behavior fails with, say, `3_`, because splitting would return `["3", ""]` and Atoi would fail to convert an empty string.

Unless you mean people should be able to post on a parent category, in which this should be fixed.

sfan5:
Oh, then that's okay already.
@akuma06 No, torrents should always have a sub-category attached.

PR #686, opened 2017-05-22
Model diagram (Ready for merge)

steindornatorinn:
Created a diagram of the models this projects uses and their interconnectivity, so nyazi shills will stop whining about lack of documention.

Marked as a WIP since I doubt I managed to do this without making a few missteps along the way.


tomleb:
can you provide a picture of what it looks like ?

steindornatorinn:
nyaapantsu_models_source

tomleb:
I don't know how I feel about this PR tbh. I feel like time would be better spent making our code self-document itself, improve it, etc than trying to keep a chart uptodate. Is the page supposed to be accessible on the website ?
I'd prefer the goal to be helping new contributors by "guiding" them in a general way, such that they can get started fairly easily (even if the code changes), like with a simple markdown file.
What do other people think ?

steindornatorinn:
The model code is split across 8 source files, no matter how well you document the code it won't replace having a diagram to give you the bigger image. The up-to-dateness of these kinds of diagrams is sort of irrelevant, there shouldn't be frequent grand changes in these after the project has come this far. Undocumented minor changes don't really matter, as this is the sort of diagram is mostly used to understand relationships, not exact reference.

This would definitely not be hosted on the webpage, the html format for the diagram just allows for easier viewing than .png.

kipukun:
I personally think this is really cool, and very useful. Good shit!

ipatrol:
I'd suggest instead of `/models`, put them in `/doc/models` so we can add other documentation in the future.

steindornatorinn:
I'm finished with it and if there are still some errors they should be minor, this is the diagram in it's current state. Moved it to a /docs directory as @ipatrol suggested too.
nyaapantsu_models

I'll commit myself to updating it once a week for the next month, this should keep it very up to date.

akuma06:
Why do you have so much files in your commit? xD

steindornatorinn:
That's what happens when you are mentally handicapped and have to use git. Fixed now.

akuma06:
Thanks! Didn't want to check every files to see what we didn't need =D

PR #688, opened 2017-05-22
Remove parent categories on upload form.

ElegantMonkey:
Parent categories aren't valid choices on a upload, so leaving them enabled is unnecessary.


sfan5:
Just ... remove them then?

ElegantMonkey:
It seems they were added when it started supporting Sukebei, I'll remove them then.

PR #690, opened 2017-05-22
New User Edit Panel

akuma06:
User can edit torrents
* delete torrents
+ New translation string for mod panel and user edit torrent panel
+ Improvement of messages util with implementation of T (no need to get
Tfunc now, messages util do that for you)
+ Use of @ElegantMonkey GetCategories to generate select of categories
in search and forms


alucard0134:
if i ever see trash like this merged here again i will revoke your push priv

PR #692, opened 2017-05-23
Revert "New User Edit Panel"

alucard0134:
Reverts NyaaPantsu/nyaa#690


akuma06:
Before Reverting, please say what is the problem? On last commit it did work perfectly and travis built correctly....

PR #695, opened 2017-05-23
Implement HEAD for views

### LINKED ISSUE #678: HEAD requests do not work on torrent page
Title.

nopjmp:
This implements HEAD methods for both /api/view/{id} and /view/{id} in a way that should only be minimal impact to the server.

Should be able to close #678 unless other HEAD requests are wanted/needed.

Let me know if something isn't quite right and I can tweak it.


yiiTT:
Works as expected

PR #698, opened 2017-05-23
Allow users to edit & delete their torrents

akuma06:
Reverts NyaaPantsu/nyaa#692


yiiTT:
@akuma06 Could you also add "Mark as trusted" for trusted users

akuma06:
@yiiTT Trusted users torrents are automatically made in trusted status.

PR #713, opened 2017-05-23
Added a link to db and IRC channels to readme

steindornatorinn:


sfan5:
Linking the database in the README is not the best idea as it could cause problems because of copyrighted content -> DMCAs.
(That link is the wrong one anyway, you'd need `merged.sqlite3.zip` instead)

steindornatorinn:
I really hate that people have to figure out that there is a link to it in the dev chat though. Would a link to a torrent be ass covering enough?

sfan5:
Just link to the dev channel maybe?

steindornatorinn:
Can't use irc links on github I'm afraid, but I'll just let people know it can be found there.

README.md
```diff
@@ -92,8 +92,12 @@ Access the website by going to [localhost:9999](http://localhost:9999).
 This project is licensed under the MIT License - see the LICENSE.md file for details
 
 # Contributing
+ Use tabs for css and js files.
 
-Use tabs for css and js files.
+# IRC
+#nyaapant<span>su@irc<span>.rizon.n</span>et</span>
+
+#nyaapantsu-dev<span>@ir<span>c.rizon.n</span>et</span> (Serious buisness only)
```
sfan5: whats up with the `<span>`'s here?

steindornatorinn: Github is retarded and tries to interpret these as email addresses, and won't allow me to actually use them as the IRC addresss that they are. It's a ugly hack but it would be really annoying to click a link  and it does something other than what you expect it to.

Alternatively I could just write on irc.rizon instead of using the @symbol, but it doesn't look as aesthetic.

PR #718, opened 2017-05-23
Fixing insert of unneeded html tags in db

akuma06:
Added a Sanitize function in util
* Possibility to add model in it
* Already a preset default model

Comments shouldn't be allowed html, too difficult to check every comment
for broken html
Torrents are still allowed html tags but I don't think it should since
we use markdown.

### DIFF
--- a/util/markdown.go
+++ b/util/markdown.go
@@ -1,12 +1,17 @@
 package util
 
 import (
+	"bytes"
+	"html/template"
+	"log"
+	"regexp"
+	"strings"
+
 	"github.com/microcosm-cc/bluemonday"
 	md "github.com/russross/blackfriday"
-
-	"html/template"
-	"strings"
+	"golang.org/x/net/html"
 )
+
 //Some default rules, plus and minus some.
 var mdOptions = 0 |
 	md.EXTENSION_AUTOLINK |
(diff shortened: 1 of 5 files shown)


yiiTT:
This is a tricky issue, I think we need to look at it a bit more, but not allowing HTML for comments.
But for the Description I don't think you should use the Strict Policy since this limits their tools the UGCPolicy should be more than enough.

yiiTT:
Suggested changes: re-implement the UGCPolicy without div tags for upload.go and then apply the custom policy on view_torrent_handler.go so we don't mess with what goes into the DB too much.

util/markdown.go
```diff
@@ -38,3 +38,24 @@ func MarkdownToHTML(markdown string) template.HTML {
...
+ * Sanitize a message passed as a string according to a setted model or allowing a set of html tags and output a string
+ */
+func Sanitize(msg string, elements ...string) string {
+	p := bluemonday.StrictPolicy()
+	if len(elements) > 0 {
+		if elements[0] == "default" { // default model
+			p.AllowElements("b", "strong", "em", "i", "u", "blockquote", "q")
+			p.AllowImages()
+			p.AllowStandardURLs()
+			p.AllowAttrs("cite").OnElements("blockquote", "q")
```
sfan5: do people ever use this?

yiiTT: yea other than cite

PR #719, opened 2017-05-23
Torrent Mass Edit Api done

akuma06:
* Torrents can be deleted in mass from frontend with api post request
* Torrents status can be edited from frontend with api post request
-- Look to function doc for more info on how to use it

It is a WIP so it might not work =D


akuma06:
Example:
`Method: Post
Url: /mod/api/torrents
Form: action=multiple&status=1&torrent_id=923009&torrent_id=923008&category=3_5`

Issue #721, opened 2017-05-23
Account theme switcher


yiiTT:
We need to implement something like the language switcher but for themes. There should be a base CSS file that is always included and also have another stylesheet depending on the theme selected.
Also should still allow people to use the quick theme switcher from day/night.

JavaScript should not be required for the theme switcher in account settings.

PR #750, opened 2017-05-25
Account theme switcher (Pls merge)

steindornatorinn:
As mentioned in #721.

I've implemented the theme switching in the same way that the languages are, (reads first from account settings, reads cookie if that's not available) so actually having this in the account settings is unnecessary since it works for visiting users as well.


tomleb:
I feel like this should be on the same page as /language and could be renamed /settings for local settings. Also should probably be added to the user setting page.

steindornatorinn:
I agree, that would make a lot of sense.

steindornatorinn:
I'm calling this ready for merge. These features are not integrated into the account settings yet but it's mostly because that page is in a bit of a disarray right now and this PR is already large enough.

akuma06:
Might need @kipukun approval before merging =)

kipukun:
All good!

PR #751, opened 2017-05-25
Replace all icons with custom version of Font Awesome

x9zb:
I used http://icomoon.io/ to generate an icon font with only the icons needed for Pantsu.
The css and the fonts only take up about ~50kb.

icons
screenshot


kipukun:
We already have them in base64, which is arguably more faster than a CSS file. But thanks!

tomleb:
We have them in base64, closing.

PR #757, opened 2017-05-26
Fix theme bugs + Reimplement "nyanpassu" and "explosion"

MMP0:


ewhal:
With the audio tags they should include preload="none" so they only load on click

MMP0:
Thanks, akuma06

akuma06:
Thanks to you for fixing the major missing feature :D "Nyaanpassu"

PR #761, opened 2017-05-26
New profile user menu

akuma06:


templates/index.html
```diff
@@ -16,13 +16,35 @@
...
+			}
+		</script>
+
     </head>
-  <body>
-     <script type="text/javascript">
-        var theme = localStorage.getItem("theme");
-        if (theme != "") document.getElementById("theme").href = "/css/" + theme;
-     </script>
+  <body onload="setTheme();">
```
MMP0: It does not work in Chrome

akuma06: Sorry for that, it was a fix for Firefox :).

MMP0: Umm... :(

PR #767, opened 2017-05-26
Activate the never_go_down again switch (fix slowdown issue)

tomleb:
Tried this in dev and on prod, seemed to eliminate the slowdown issue.
We were having a lot of 'idle' and 'idle in transaction' connection issue with pgpool.
Connection should never go idle now which should let new connections in.


tomleb:
Setting SetMaxIdleConns(-1) is the culprit, will try to work around it. (I don't know why it fails with sqlite)

PR #768, opened 2017-05-26
Trackers in Torrents + Missing comments + Function renaming

akuma06:
* Added some missing comments
* Renamed functions to get user followers/following
* GetFollowers to get followers
* GetLikings to get who the user is following

### DIFF
--- a/config/trackers.go
+++ b/config/trackers.go
@@ -13,3 +13,8 @@ var Trackers = []string{
 	"udp://tracker.internetwarriors.net:1337/announce",
 	"http://mgtracker.org:6969/announce",
 	"http://tracker.baka-sub.cf/announce"}
+
+// NeededTrackers : Array indexes of Trackers for needed tracker in a torrent file
+var NeededTrackers = []int{
+	0,
+}
--- a/model/torrent.go
+++ b/model/torrent.go
@@ -1,13 +1,15 @@
 package model
 
 import (
+	"context"
 	"fmt"
 	"html/template"
-	"context"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
+
 	elastic "gopkg.in/olivere/elastic.v5"
 
 	"github.com/NyaaPantsu/nyaa/config"
@@ -52,6 +54,7 @@ type Torrent struct {
 	Filesize    int64     `gorm:"column:filesize"`
 	Description string    `gorm:"column:description"`
 	WebsiteLink string    `gorm:"column:website_link"`
+	Trackers    string    `gorm:"column:trackers"`
 	DeletedAt   *time.Time
 
 	Uploader    *User        `gorm:"AssociationForeignKey:UploaderID;ForeignKey:user_id"`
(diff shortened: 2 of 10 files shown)


config/trackers.go
```diff
@@ -13,3 +13,8 @@ var Trackers = []string{
 	"udp://tracker.internetwarriors.net:1337/announce",
 	"http://mgtracker.org:6969/announce",
 	"http://tracker.baka-sub.cf/announce"}
+
+// NeededTrackers : Array indexes of Trackers for needed tracker in a torrent file
```
sfan5: why should this tracker be required?

model/torrent.go
```diff
@@ -67,27 +70,9 @@ type Torrent struct {
 }
 
 // Size : Returns the total size of memory recursively allocated for this struct
-// FIXME: doesn't go have sizeof or something nicer for this?
+// FIXME: Is it deprecated?
```
sfan5: it's actually used by `make()` to preallocate memory as far as i know

config/trackers.go
akuma06: Dunno, that's our tracker and it was decided to implement it in our torrents?
I think there was an issue and then they added the information on upload page

service/upload/upload.go
```diff
@@ -1,12 +1,15 @@
...
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/NyaaPantsu/nyaa/config"
 	"github.com/NyaaPantsu/nyaa/model"
 )
 
+const trackerRegex = `^(http[s]*|udp)://[a-z0-9\.:\-/?]+$`
```
sfan5: tracker urls can contain `&`, `=`, `[` and `]` too
why are you even validating this?

config/trackers.go
sfan5: Yes but it seems having the tracker is <i>required</i>, not <i>optional</i>

service/upload/upload.go
akuma06: Can't you have an attempt on inserting other data in the db by falsifying a .torrent file?
For torrent tracker url, I just tried the regex with the ones in our config.
If you think we don't need to be sure that it is a tracker url, then I can just remove the regex :)

config/trackers.go
akuma06: When I read this: "Please include udp://tracker.doko.moe:6969 in your trackers.", I get that this is not optional but required. Or we should change it to: "You may add xxx"

akuma06: But if they are not in a torrent, it doesn't show an error, it just adds it to the existing trackers of the torrent.

service/upload/upload.go
akuma06: Are you sure about that? I'm not quite knowledgeable on torrents tracker url but on the list that I've I don't see the characters you provided :/ (I just googled: list of trackers url)

config/trackers.go
sfan5: oh, that's okay then

akuma06: No that's okay. Good to ask :). Can you reply to me for the tracker regex? I don't know if I should check for "&" or not. If I have to, we need to convert "&" characters then because how do magnet will work?

service/upload/upload.go
sfan5: How are trackers stored in the db? If done safely then this regex should not be needed

Here's an example: `http://[2001:db8::1]:6969/announce?a=1&b=1?`

akuma06: Ok thanks :/. In magnet urls you encode the url of the tracker then? Because the "&" character would not work if not encoded?

PR #794, opened 2017-05-27
Some SEO and SMO works

inpothet:
Done some reworks of the SEO and SMO
PS we should get a 120x120 logo that can be used for the SMO as i dont know if google like me using the Favicon for now


inpothet:
Well this is fun wtf why does it not compile if the title is used 2 times
it was needed somewhere els for the SMO to

steindornatorinn:
The block tag does not do what you think it does. From the documentation:
>{{block "name" pipeline}} T1 {{end}}
>	A block is shorthand for defining a template
>		{{define "name"}} T1 {{end}}
>	and then executing it in place
>		{{template "name" .}}
>	The typical use is to define a set of root templates that are
>	then customized by redefining the block templates within.

Change the second and third uses to template and it should work.

templates/index.html
```diff
@@ -2,24 +2,32 @@
...
-		<link rel="icon" type="image/png" href="/img/favicon.png?v=3" />
-
+    <!-- Flavicons and icons -->
+    <link rel="icon" type="image/png" href="/img/favicon.png" />
+    <link rel="shortcut icon" type="image/png" href="/img/favicon.png">
+    
+    <!-- SMO -->
+    <meta property="og:site_name" content="Nyaa Pantsu">
+    <meta property="og:site_name" content="{{if Sukebei}}Sukebei{{else}}Nyaa{{end}} Pantsu - {{template "title" .}}{{ call $.T "error_404" }}{{end}}">
+    <meta property="og:image" content="/img/flavicon.png">
```
sfan5: typo in url

alucard0134:
why did i have to fix this myself

PR #796, opened 2017-05-27
Reimplemented night mode and mascot disabling(don't merge)

steindornatorinn:
This is all JS, but the actual themes behind this can be used from the settings menu without it.


public/js/main.js
```diff
@@ -11,9 +11,23 @@ function toggleNightMode() {
...
+	theme = opt.value;
+	localStorage.setItem("theme", theme);
+	document.getElementById("theme").href = "/css/" + theme;
+	console.log(theme);
+}
+
+*/
+
+
+function switchThemes(themeName=null){
```
cfrank: I'm not sure about what happens when a browser that doesn't support default arguments encounters them. Is is a graceful failure or does it throw an error? I don't have access to any computers with IE, can you test this?

steindornatorinn: Unfortunately no, I'm running on a virtual machine with a wifi connection and I can't bridge it over to windows because drivers. 
I'll just replace it with the older syntax so I can be sure it works.

steindornatorinn:
I'm closing this one and opening another one due to some weird merge conflict I had trouble resolving, ended up just moving my changes to a fresh branch.

Issue #736, opened 2017-05-25
Torrent Title instead of Hash for Torrent File Name.


printabcx:
For organization sake, will there be plans to change the naming system for torrent files to the title of the torrent rather than naming it with it's hash?

loadletter:
It could be easily done by adding a download attribute to the link, but it won't work on Firefox because torrents are stored on a different domain.

printabcx:
> It could be easily done by adding a download attribute to the link, but it won't work on Firefox because torrents are stored on a different domain.

Not sure what it means. I'm just an ordinary user, so I don't know jargon related to developers. Could you elaborate on the attribute thing with the link with Firefox?

loadletter:
I'm not a developer, it's just a suggestion because I had to deal with it, basically you can make a download link on a page download with a different name (which means no changes needed to existing torrents or torrent serving infrastructure) but firefox disallows this if the page with the link and the .torrent file are on different domains because of security implications.

PR #800, opened 2017-05-27
Torrent name when download

akuma06:
New Download Torrent controller when dowloading a torrent
Route download need to be set the same as in TorrentStorageLink


akuma06:
Should fix #736

PR #811, opened 2017-05-28
Fix "see more torrents from" box width

Kiloutre:
The width is fixed to 35px because all \<li\> are fixed at 35px width, which is obviously not enough for the text it is trying to contain.
At first i tried to remove that 35px and just let the size change based on nickname length, but making it stick to 100% looks better in my opinion

Here is how it looks right now:
picture1
Here is how it would look without the 35px:
picture2
Here is how it would look with my change to force it at 100% width:
picture3


tomleb:
It seems to be already fixed on [dev](https://dev.pantsu.cat/user/5199/QQQ)

PR #816, opened 2017-05-28
Removal of </img>

Kiloutre:
Removal of \</img\> because it does not exist
Addition of / at the end of  because of consistency

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -44,7 +44,7 @@
 		<div class="container">
 			<div class="h-left">
 					<div class="h-logo">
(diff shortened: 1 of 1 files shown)


tomleb:
According to the examples on [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img], it doesn't need the '\' at the end.

Kiloutre:
> it doesn't need the "\\" at the end.

This is true but kipukun told me to put it for consistency, because \<br/\> has it (and it doesn't need it either)

PR #820, opened 2017-05-28
Fixing the mascot

Duckbitts:
Added a class for the "Mascot" button. Added some pointer-events to make the "Mascot" button clickable when the window is narrow enough to make the mascot image overlap the button.


tomleb:
Won't that make impossible to click anything though ? Having pointer-event: none on  #content

Duckbitts:
Yes, but I added pointer-event: auto; into things that should be clickable. That overrides pointer-events: none; and makes those clickable again.

akuma06:
What is the bug there? I don't understand the need for this? Can you explain with more details?

Duckbitts:
There are divs overlapping the image. Those divs are making the mascot unclickable except for a small spot.
Pointer-events have been used to make the user able to click through the divs.
Do ask if my explanation is lacking.

PR #821, opened 2017-05-28
Add playbook to download /build latest nyaa, fix k-on elasticsearch issue

tomleb:
The playbook is a simply utility for when we deploy on dev (I wouldn't use it for prod yet because it doesn't take into account sukebei)
This setting of elasticsearch gives better result (should still be tested more heavily than my ~2-3 search)

### DIFF
--- a/deploy/ansible/roles/elasticsearch/files/elasticsearch_settings.yml
+++ b/deploy/ansible/roles/elasticsearch/files/elasticsearch_settings.yml
@@ -2,19 +2,41 @@
 settings:
   analysis:
     analyzer:
-      nyaapantsu_analyzer:
-        tokenizer: nyaapantsu_tokenizer
+      nyaapantsu_search_analyzer:
+        type: custom
+        tokenizer: standard
+        char_filter:
+          - my_char_filter
         filter:
           - standard
           - lowercase
-    tokenizer:
-      nyaapantsu_tokenizer:
-        type: edge_ngram
+      nyaapantsu_index_analyzer:
+        type: custom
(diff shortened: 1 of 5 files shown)


deploy/ansible/roles/elasticsearch/files/elasticsearch_settings.yml
```diff
@@ -2,19 +2,41 @@
...
+        type: pattern_capture
+        patterns: ["(\\d+)[xX](\\d+)"]
+      word_delimit:
+        type: word_delimiter
+        preserve_original: true
+        split_on_numerics: false
+    char_filter:
+      my_char_filter:
+        type: mapping
+        mappings: ["-=>_", "!=>_", "_=>\\u0020"]
```
Atvaark: Since this is the ES config used by nyaa.si, doesn't that mean you'll have to switch the license to GPL?

tomleb: No, MIT is compatible with GPL afaik

Nazonaze: MIT is only compatible with GPL in that you can use MIT code in a GPL project with attribution, but the same is not true the other way around because MIT doesn't have the same FOSS guarantees that GPL does, thus relicensing would be necessary.

And proper attribution would be required regardless, which this is also lacking.

ewhal: Add a license header to the config file

PR #837, opened 2017-05-29
CSRF Support + better key for context

akuma06:
* Added new dep: gorilla/csrf
* CSRF field in forms
* CSRF variable in commontemplatevariables
* New key for messages and user context

Please change EnableSecureCSRF to false when testing locally and don't
merge config/env.go with the changes


main.go
```diff
@@ -29,8 +30,13 @@ func RunServer(conf *config.Config) {
 	os.Mkdir(router.DatabaseDumpPath, 700)
 	// TODO Use config from cli
 	os.Mkdir(router.GPGPublicKeyPath, 700)
-	http.Handle("/", router.Router)
 
+	// Please make EnableSecureCSRF to false when testing locally
+	if config.EnableSecureCSRF {
+		http.Handle("/", csrf.Protect([]byte("q8satbudwexfzh2j3m5n6p8r9satcvsd"))(router.Router))
+	} else {
```
Atvaark: Shouldn't this be stored in a config file or env variable instead of being hardcoded and public?

akuma06: It should have been :). Forgot to remove the test token

akuma06:
Should fix #831

PR #843, opened 2017-05-29
Playbook to install latest nyaa straight from github

tomleb:
Was reverted because of the whole es config drama.


tomleb:
Before merging: The current behavior is that if some changes were made in the directory of the build previously and you're trying to run this script, it won't work.
I can add force: yes to ansible, but that will erase all changes that were made.
Which behavior do we want ?
I could also modify it to make a backup of the current folder before getting the new one.

akuma06:
3rd, then 2nd and last 1st I would say

tomleb:
# Don't merge yet
I'll modify it to make a backup of the current folder before pulling latest.

tomleb:
I make a backup of the folder now before pulling the latest.

# Ready to be merged

PR #854, opened 2017-05-30
New config files

akuma06:
As decided, config files are parsed at runtime.
I decided to go for YAML config files because there can be comments in
it.
There are 2 files:
* config/default_config.yml <= which shouldn't be edited unless we add a
config parameter
* config/config.yml <= which is the user-defined config. This file
shouldn't be commited

Changed every call to config.XXX to config.Conf.XXX (look to the new
stucture of config in config/types.go)

Of course, putting config parameters in config.yml overrides config in
config_default.yml. You don't have to put everything in it, just add
what you want to override.


config/default_config.yml
```diff
@@ -0,0 +1,164 @@
...
+# database type
+db_type: sqlite3
+# DBParams will be directly passed to Gorm, and its internal structure depends on the dialect for each db type
+db_params: ./nyaa.db?cache_size=50
+# logmode for database
+db_logmode: default
+# Environment should be one of: DEVELOPMENT, TEST, PRODUCTION
+environment: DEVELOPMENT
+# WebAddress : url of the website
+web_address: nyaa.pantsu.cat
```
Atvaark: Maybe use *localhost* (without HTTPS) as a default website value and set the actual website in config.yml during deployment?

Otherwise every developer would have to be careful not to add their *config.yml* changes to their commits.

akuma06: config.yml should be in .gitignore maybe?

akuma06: In fact config.yml shouldn't be commited, they are user-specific configs.

Atvaark: .gitignore only works for untracked files. Since the example config.yml is tracked every change to it will be listed by *git status* and be staged to commit by e. g. *git add .*.

You could maybe add config.yml to .gitignore add not provide an example.
But I'd prefer sane defaults and not having to fiddle with a config file to test a local change.

akuma06: Yeah forgot to rename it in config_example.yml

config/config.go
```diff
@@ -91,22 +34,35 @@ var allowedDBLogModes = map[string]bool{
...
+// Construct a new config variable
+func init() {
+	getDefaultConfig()
+	privateConf = *DefaultConfig
+	Conf = &privateConf
+	overrideDefaults()
+}
+
+func overrideDefaults() {
+	path := "config/config.yml"
```
Atvaark: This'll cause issues in every unit test that imports this package. (Except for the main package,)

Unit tests don't use the project root as the working directory and will fail to find these files.
(The init function is run before the test functions)

You'll have to make sure the default config file is always found or the tests will fail in strange ways.

akuma06: Thanks again @Atvaark, you're the only one that helped :')

PR #857, opened 2017-05-30
Maskot to cookie

steindornatorinn:
Added the maskot hiding to the cookie as well, leaving none of the configuration up to JS.


mitki:
Isn't {{call $.T "mascot_select"}} redundant? Since there's {{call $.T "mascot"}} already.

steindornatorinn:
Yes it is, I'll remove it.

PR #865, opened 2017-05-31
Update URL queries

MMP0:


tomleb:
This should be done at runtime instead, generating a hash when the application starts and on -SIGHUP.

akuma06:
Just use {{ .Conf.Build }} to get the build version

akuma06:
Or if you just want the version number {{ .Conf.Version }}. Be aware that for {{ .Conf.Build }} to be different from unknown, you need to use the go build command in readme

PR #895, opened 2017-06-03
New modpanel fix

### LINKED ISSUE #755: Redesign / ES new dev issues
I'll list the issues people find on the dev environment here:
- ~~ES search sorts on ID's by default (Looks like postgresql search sorts by date by default, to investigate)~~
- ~~No more explosssiiiioonnnnn (need to be added)~~ (#757) 
- ~~Wrong color of alternating row when the torrent is APlus, Trusted, etc.~~

- ~~Mobile needs a search bar~~
- ~~Theme defaults back when searching or changing order. (Whenever we load a new page)~~
- ~~Search button becomes invisible on tomorrow.~~
- ~~"The buttons aren't consistent widths. E.g. The "1" button is thinner than the "10" button. Which makes them crammed together and look janky." (Should probably make a fixed width for the pagination button)~~
- ~~Above wanted, current is bottom [pic](https://my.mixtape.moe/ugjjlv.png)~~
- Search puts a '*' when empty search
- Search box size and font size is small
- Date and time in torrent info has the time cut off due to the short column.
- Need wrapping of long character comments
- Icons are very hard to see on mobile using the /g/ theme.
- Z-index on the mascots is blocking the menu in small width views
- Edit torrent button is missing
- ES search seems to have worse performance than my 4gb machine :') (Getting queries of 70 ms instead of ~9 ms)
- Make /g/ Anon default theme (discuss)
- Duplicated comments See [https://nyaa.pantsu.cat/view/865002](https://nyaa.pantsu.cat/view/865002)
- Improvement to torrent view ?
- Duplication of lots of torrents (Might be ES related ?) (Multiple entries in db confirmed, need to check upload verification asap)
- "Invalid date" in comments

akuma06:
Fixes #755 (@sfan5)
modpanel tables have margins
modpanel delete buttons are red
some coloring for buttons
reassign form looks better
reassign form textarea have rows="20" cols="40"
"save changes" btn green
report reasons are now translated in modpanel
delete is a real button on modpanel users page


sfan5:
@kipukun might want to look at the style changes too

templates/view.html
```diff
@@ -75,10 +75,10 @@
                 <b>{{ call $.T "report_type" }}:</b>
                 <form method="post" action="/report/{{$.Torrent.ID}}">
                     {{ block "csrf_field" $ }}{{end}}
-                    <input type="radio" name="report_type" value="illegal" id="illegal" required> <label for="illegal">{{ call $.T "illegal_content" }}</label><br />
-                    <input type="radio" name="report_type" value="spam" id="spam" required> <label for="spam">{{ call $.T "spam_garbage" }}</label><br />
-                    <input type="radio" name="report_type" value="wrongcat" id="wrongcat" required> <label for="wrongcat">{{ call $.T "wrong_category" }}</label><br />
-                    <input type="radio" name="report_type" value="dup" id="dup" required> <label for="dup">{{ call $.T "duplicate_deprecated" }}</label><br />
+                    <input type="radio" name="report_type" value="illegal_content" id="illegal" required> <label for="illegal">{{ call $.T "illegal_content" }}</label><br />
+                    <input type="radio" name="report_type" value="spam_garbage" id="spam" required> <label for="spam">{{ call $.T "spam_garbage" }}</label><br />
```
sfan5: you still need to rename these back

akuma06: Why would I need to do that?

akuma06: The goal is to move in the future (when old reports have been read and deleted) to just translate, instead of using the function.
1. This would mean that reports will be translated
2. There will be only one translate string per item (not two)

So for now we have backward compatibility.

sfan5: wait nevermind
`getReportDescription` should call T() with "illegal_content"

akuma06: We don't need that :). illegal_content is a string in the translation. So why would you call getReportDescription? 
getReportDescription has one and only use: the report listing on torrent_report.html. Why? Because the value of the old reports are not the same as the one in this form. So when you try to call $.T, it can't translate those. However, getReportDescription will change the old values

akuma06: Anyway, if you've tested it you should see what I mean and how it works. getReportDescription is just a function that will disappear because it's a fix for old report description in the database

Issue #902, opened 2017-06-03
Logout should require token


AnthonyFoley:
I can embed images in comments and torrent descriptions which have `https://dev.pantsu.cat/user/logout` as the target. Upon viewing this torrent or comment, the user is then logged out.

Markdown code: `![](https://dev.pantsu.cat/user/logout)`

Possible solutions: 
- Require some kind of token parameter for the logout URL
- Disable image embedding for `/user/logout` as both relative and full and urlencoded URLs (maybe more edge cases like URL shorteners, who knows)

PR #921, opened 2017-06-04
add CSRF protection to logout

gjwufq3x:
Fixes issue #902


templates/_badgemenu.html
```diff
@@ -11,7 +11,7 @@
 					<a class="nav-btn" href="{{ genRoute "user_notifications" }}">{{ call $.T "my_notifications"}} <span class="badge">({{ .GetUnreadNotifications }})</span></a>
 					<a class="nav-btn" href="{{ genRoute "user_profile_edit" "id" (print .ID) "username" .Username }}">{{call $.T "settings"}}</a>
 					{{if HasAdmin . }}<a class="nav-btn" href="{{ genRoute "mod_index" }}">{{call $.T "moderation"}}</a>{{end}}
-					<a class="nav-btn" href="{{ genRoute "user_logout" }}">{{ call $.T "sign_out"}}</a></li>
+					<a class="nav-btn" href="{{ genRoute "user_logout" }}?token={{ $.LogoutToken }}">{{ call $.T "sign_out"}}</a></li>
```
Atvaark: Why not use a form with **method="post"** and the gorilla **csrf_field** field?

gjwufq3x: If tokens from gorilla don't expire (I haven't looked at that project) that could work too. I'm mostly concerned about it throwing errors or inconveniencing people that leave a page open for a while.

PR #948, opened 2017-06-06
Fix translation IDs

MMP0:


Myrmece:
Oh right I forgot to fix those ones, thank you.

PR #970, opened 2017-06-07
Country flags (language) for torrents.

ElegantMonkey:
#964

Show country flags for torrents, both on Home/Search and View Torrent page. Added a new field, `.Language`, to the Torrent model, and a `torrentLanguages` package which reads the available languages and flags from the translation files. Also enforces that English categories (English-translated anime, literature and live action) only have English torrents, and non-English ones have non-English torrents.

A few points:
- Language is required for all new uploaded torrents (I don't know if it should be only enforced for "non-English translated" categories, but on Anidex it seems all torrents have it, so...)
- Old torrents won't show any flags. Maybe it's possible to show an English flag for torrents under English categories, but some torrents having a flag and others not is kinda weird IMO.
- The flag pack used right now is from [googlei18n/region-flags](https://github.com/googlei18n/region-flags), which consists of (mostly) public-domain ones. The images were resized to 128x128.


akuma06:
Would need approval of @kipukun and @ewhal for this :). Seems good to me (haven't tested).
However, if the language is not supported by the website (doesn't have any translation for it), won't that forbid people from choosing the right language?

sfan5:
How do you even "read" languages from a torrent file?

also country flags are <i>the</i> worst way to approach this:
* do you display all of `US` `GB` `AU` `NZ` IE` at once for English?
* does `BE` mean French or Dutch?
* does `ZA` mean Afrikaans, English or any of the 9 other official languages of south africa?
* how do `DE` `CH` `AT` or `LI` differ? (do they?)
* does `AD` refer to Spanish, French or Portugese?
* how do you differ between cyrillic and latin Serbian?
* is `LU` French or German? or is it their local language?
* which language is `EU`?
* is `SG` English, Malay or Mandarin (Chinese)?
* are `FR` and `RE` the same thing? are `SE` and `AX` the same?

not to mention that country flags look ugly af

ElegantMonkey:
@akuma06 Right now, yeah, it wouldn't support it. I thought about it, but I couldn't find an approach that didn't involve hardcoding flags and languages somewhere in the code or in other somewhat weird way.

@sfan5 When I said "read the available languages" I meant reading them from the translations (ie. what languages do we support), not reading from the torrent itself, sorry if I wasn't clear. The uploader would choose it when uploading. Right now, all our translations are in the format `[language code]-[country code]`, and that's where it gets which flag to use (this can be overriden by setting a "flag" translation inside the translation file, like on the `en-us` one, which uses the `GB` flag).

> not to mention that country flags look ugly af

:^)

kipukun:
A much, much better alternative would be to use this service: https://www.flag-sprites.com
We only generate the sprites we need for each language we support. 
`{{.Language}}`
Something of that nature. It would decrease load times and it looks much better than the current flags.

sfan5:
An even much more better approach would be to not use country flags at all because those don't map to languages 1:1

kipukun:
I don't really see a point if we already categorize torrents as non-English and English. That should probably be good enough, as @sfan5 is right.

ewhal:
We have a lot of varying languages for non-english translated.
This makes life difficult for non-english and non-japanese users.

ElegantMonkey:
@kipukun Will take a look at flag sprites, thanks. Also, I think @ewhal is right, it's really hard to search for something other than English and Japanese, especially when many fansubs don't write the language on the torrent title.

@sfan5, while I do agree that they won't map 1:1, it's the simplest way of showing them in a somewhat friendly way. Do you have any suggestions?

sfan5:
My suggestion is to just have some pre-selectable languages, then display them as text and allow them to be searched for.

ElegantMonkey:
If everyone is okay with this, I'll do it then (although I'm not sure how to make it searcheable, wouldn't this be on ElasticSearch or something?)

akuma06:
For the search, we would need @kipukun to add the advanced fields on the search form (dates, filesizes, ...). So it can wait until they are added.

I understand the point of @sfan5, however flags are still a good way to display the actual language without overcharging the torrent list table. If we put the actual name of the language, it will reduce considerably the space for other stuff (and we are still lacking some space like the bug with fulldates on foreign languages taking too much space).

So if we are doing that, they are not that many solution:
1. Make the .box larger and hiding some infos on little screen: will resolve the bug with foreign dates (additionally make the date a short one on little screen, like we did previously for comments), will be able to add full language string (additionally reduce it to language code or flag when little screen)
2. Put the flag on torrent view (easier to find torrents with flag color) and already used by many translation tools (reverso, google translate for websites, ...)
3. Put the language code (not everyone understand it)
4. Put the language string and fuck it those who finds that there is not enough space!

ElegantMonkey:
@tomleb, is this okay? Or will I need to add it somewhere else?

As there doesn't seem to have a consensus on how to display the language, maybe we shouldn't show it on the search/home listing, and only show the text on the Torrent Details page? Being able to search by language would already be a nice feature, IMO.

tomleb:
Looks fine to me, but you'll need to add it to https://github.com/NyaaPantsu/nyaa/blob/dev/model/torrent.go#L236 because when getting an entry from ES we get it as model.TorrentJSON and we then convert it to model.Torrent. (Btw, this whole thing needs to change, we should have only one representation of a Torrent)
Also if you want to enable sorting by language you'll have to add fielddata: true (look at the torrent_name example in the ES configuration)

ElegantMonkey:
Ok, the field is already added to `.ToTorrent()`, but I can't find this `fielddata: true` you mention in the ES settings.

akuma06:
@ElegantMonkey Look there he seems to talk about that: https://github.com/ElegantMonkey/nyaa/commit/ba683c3bcbc8c64680f3a7ab5043c646b64c8a7d#diff-a02858613a257740b553cc96d8939edeL32

tomleb:
Ah right, I forgot I made that change.
So basically when indexing 'text', the entry gets processed by the analyzer whereas keyword leave the content untouched.
The fielddata: tue is needed to sort by text but it's not the case for keyword.
So change the field type to keyword in this case.

tomleb:
You don't need the raw thing, that's only for multiple indexing (like the torrent_name is indexed both by being processed by the analyzer and as raw keyword.)
We only want to store the language as keyword.
```
       language:
         type: keyword
```

ElegantMonkey:
@tomleb, from the ES site:

>Keyword fields are only searchable by their exact value.

However, at least in the non-ES implementation, the languages are searcheable by prefix/suffix (eg. `pt` would match both `pt-pt` and `pt-br`). I don't think this is necessary, but it's a nice thing to have.

tomleb:
You could also just use wildcards for the search and still store using keyword.
So you could do pt* and find pt-br and pt-pt.
For your edit, yeah you could do that. You could search with the field language and sort with the field language.raw (Do we even need sorting ?)

ElegantMonkey:
> Do we even need sorting?

Probably not, I'll add the `simple` analyzer and keep it on a `text` field, which seems to be the best solution.

tomleb:
Can we rename Language to SubLanguage or something like that ? Just language seems confusing if it's for subtitles.

ElegantMonkey:
It can be used for other things, like raws, dubs or manga TLs.

tomleb:
Ah ok, can we make a comment as to what it is then in the code ? Just something like 'can be used for other things, like raws, dubs or manga TLs.' would be enough.

ElegantMonkey:
Is this ok?

tomleb:
Yes thank you

tomleb:
Tested manually with this config:
```yaml
---
settings:
  index:
    number_of_shards: 1
    number_of_replicas: 0
    max_result_window: 30000

mappings:
  torrents:
    properties:
      language:
        type: text
        analyzer: simple
```

```bash
[vagrant@localhost ~]$ curl -XPUT 'http://localhost:9200/language_test/torrents/1?pretty' -d '{"language":"pt-br"}'
[vagrant@localhost ~]$ curl -XPUT 'http://localhost:9200/language_test/torrents/2?pretty' -d '{"language":"pt-pt"}'
[vagrant@localhost ~]$ curl -XPUT 'http://localhost:9200/language_test/torrents/3?pretty' -d '{"language":"en-us"}'
[vagrant@localhost ~]$ curl 'http://localhost:9200/language_test/_search?pretty&q=language:pt'
{
  "took" : 132,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "failed" : 0
  },
  "hits" : {
    "total" : 2,
    "max_score" : 0.59907764,
    "hits" : [
      {
        "_index" : "language_test",
        "_type" : "torrents",
        "_id" : "2",
        "_score" : 0.59907764,
        "_source" : {
          "language" : "pt-pt"
        }
      },
      {
        "_index" : "language_test",
        "_type" : "torrents",
        "_id" : "1",
        "_score" : 0.42169982,
        "_source" : {
          "language" : "pt-br"
        }
      }
    ]
  }
}

```
If that's result that you want then all good.

ElegantMonkey:
Yeah, that's what I wanted. Thanks!

ewhal:
ok to merge?

PR #1004, opened 2017-06-15
Feed eztv + torznab

akuma06:


ewhal:
Just needs the t=caps endpoint I believe

PR #1013, opened 2017-06-16
Fix chrome store link

MMP0:


akuma06:
Completely forgot to remove that :D

PR #1059, opened 2017-06-23
Make flags overlap category icon

Kiloutre:
It should do this:

ss

I haven't tested it in go & js, so it could possibly not work. The CSS and HTML however are tested and do work, but the GO and JS generating the HTML might not work properly so someone will have to test that.

### DIFF
--- a/public/css/flags/flags.css
+++ b/public/css/flags/flags.css
@@ -7,6 +7,9 @@
     width: 16px;
     height: 11px;
     background: url('flags.png') no-repeat;
+	position: relative;
+	left: -22px;
+	bottom: -22px;
 }
 
 
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -274,9 +274,9 @@ th, .home-td {
 }
 .tr-cat .nyaa-cat img {
 	position: absolute;
-    bottom: -6px;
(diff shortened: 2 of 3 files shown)


sfan5:
it's <i>supposed</i> to overlap?
what kind of retarded design choice is that

Kiloutre:
Right now it's like that:

flag

So it overlapping is definitely better than the mess it is right now in my opinion
One step forward!

sfan5:
why not just put it in a separate (but small) column to the right of the category?

Kiloutre:
You could if you want but i like it better that way

Kiloutre:
ss 2017-06-23 at 11 06 19
Those are the changes akuma06 just made. I disagree with the flag position because the flag isn't even being shown fully (bottom pixel line missing) and more importantly the flag is to the right just like the text, which the left side of the picture is empty. Having this many information in such a small space is cluttering in my opinion, having the flag at the left also allows it to be less eyecatching if you're not looking for flags and just looking at the torrent title.

Kiloutre:
411a34961d043c8bc87895b9343445fa

Looks like that but one pixel higher and one pixel more to the left.

PR #1078, opened 2017-06-26
First batch of changes for the refactor

akuma06:
Added the support of gin in routes and other services/utils
Begining implementation of JetHTML


service/user/form/form_validator.go
```diff
@@ -1,42 +1,26 @@
...
 // EmailValidation : Check if an email is valid
 func EmailValidation(email string, mes *msg.Messages) bool {
-	exp, errorRegex := regexp.Compile(emailRegex)
-	if regexpCompiled := log.CheckError(errorRegex); regexpCompiled {
-		if exp.MatchString(email) {
-			return true
-		}
-		mes.AddErrorT("email", "email_not_valid")
-		return false
+	if govalidator.IsEmail(email) {
```
akuma06: Need to be reverted the regexp from govalidator doesn't valid all emails!
And consider some invalid pattern as valid emails.

PR #1085, opened 2017-06-28
Remove footer's top margin

Kiloutre:
Because why is such a big amount of spacing needed?

Before:

before

After:

after


Kiloutre:
why u clos me ;-;

PR #1096, opened 2017-07-03
Flag clickeable to get to flag search

Kiloutre:
Clicking a flag will get you to /search?c=[category]&f=[flagcode]. The <a> for this is a wide and tall as the flag (16px 11px).
The code for flag search hasn't been implemented yet though so that's a preliminary change.

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -264,19 +264,17 @@ th, .home-td {
 	position: relative;
 	text-align: left;
 }
-.tr-cat a { 
+.tr-cat a.category { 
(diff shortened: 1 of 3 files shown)


Kiloutre:
Some changes were made in search-refiner but the checkbox each have a different input id so in the GET parameters it'll look like 

> search?c=4_4&flag-fr=1&flag-us=1&flag-br=1

I want it to be like 

> search?c=4_4&f=fr_us_br

But i don't have the skills to do that so someone help pls

PR #1117, opened 2017-07-07
Make categories borderless and add border through css

Kiloutre:
Allows for flexibility in the categories picture width (smaller on phone?) but retains the same look as before on desktop. Those changes allow the .tr-cat to be down to 77px minimum (90px right now) and the .nyaa-cat (category icon) to 68px (down from 80px). Beyond that, LiveAction/Sub will not fit in the icon.

Before (Desktop):
before

After (Desktop):
after
(Yeah, it looks the same, but if you zoom a lot on the categories icon you can see an extra pixel on the before.png)

Minimum-size (Desktop):
saving space

Before (Phone):
phone before

After (Phone, 810px wide and downwards):
phone after

It's switching to reduced size once it hits 810px. I'd like the .tr-cat \<td\> to change the size on it's own though, as in having a defined minimum & maximum width and let it swim in these limits all he wants, Reducing it's size allows for the category image to be cropped on it's own.


akuma06:
Can't use it, make it to refactor please :')

Issue #915, opened 2017-06-04
Fap link disappears in some languages


MMP0:
![tes](https://cloud.githubusercontent.com/assets/28616020/26763392/2997ef30-498d-11e7-9c58-725e0e0f49fe.PNG)

steindornatorinn:
The problem here is just that the header is overcrowded at the moment.

I want to do one ore more of the following:
Replace the RSS link with the RSS button at all window sizes.
Put the results per page somewhere else, make it a cookie setting or some sort of "advanced search" feature. It doesn't really belong on the navbar.
Remove avatars from the navbar, they aren't really needed there.
Move the FAQ somewhere else, since it is the biggest source of language settings breaking the navbar. It's really hard to deal with a 3 letter word becoming something like "Preguntas frequentes" when we are already starved for space.

akuma06:
The navbar has more space now, so it is no more an issue

PR #1119, opened 2017-07-08
Fix nav width

MMP0:


akuma06:
Seems like it makes the header taking the full width of the screen

public/css/main.css
```diff
@@ -52,7 +52,7 @@ body {
 
 .header .container {
 	position: relative;
-	width: auto;
+	max-width: none;
```
akuma06: With this, you remove the max-width of the header. So it takes the whole space and so its width is equal to the width of the screen

MMP0:
Some links disappears again, like this: #915

akuma06:
I know but you people doesn't want a wide header. Might be better to move FAQ link in the footer instead

MMP0:
I see

PR #1121, opened 2017-07-08
Update Master from dev

akuma06:


akuma06:
Too much conflicts... We should just remove master as @ewhal said and create it again

PR #1124, opened 2017-07-08
Remove flags showing up on Sukebei

Kiloutre:
In the past month, as people were forced to upload with flags (remember i did a commit for that but it's still not on prod) which caused a fuckton of people uploading with random flags
And the vast majority of torrents on Sukebei are in japanese so it's not really an issue

sukebei


ewhal:
Sounds like we need to make no flag default or something else.
I don't agree with removing flags entirely though

PR #1156, opened 2017-07-11
[wip] Add catalan flag

Kiloutre:
When zoomed in it almost look as good as the others
catalan
Still need to change catalan flagcode though


akuma06:
Merged in #1158

PR #1175, opened 2017-07-14
making comment column invisible if no comment & some other things

Kiloutre:


templates/site/torrents/listing.jet.html
```diff
@@ -55,14 +55,16 @@
                   </a>
               </div>
               </td>
-              <td class="tr-name home-td">
+              <td class="tr-name home-td"{{ if len(.Comments) > 0 }}colspan="2"{{ end }}>
```
akuma06: should be == 0

public/js/main.js
```diff
@@ -83,6 +83,8 @@ document.getElementsByClassName("form-input refine")[0].addEventListener("click"
     document.getElementsByClassName("box refine")[0].style.display = document.getElementsByClassName("box refine")[0].style.display == "none" ? "block" : "none";
     if(document.getElementsByClassName("form-input refine-searchbox")[0].value != document.getElementsByClassName("form-input search-box")[0].value)
         document.getElementsByClassName("form-input refine-searchbox")[0].value = document.getElementsByClassName("form-input search-box")[0].value;
-    e.preventDefault();
+    if(document.getElementsByClassName("form-input refine-category")[0].value != document.getElementsByClassName("form-input form-category")[0].value)
+        document.getElementsByClassName("form-input refine-category")[0].value = document.getElementsByClassName("form-input form-category")[0].value;
```
akuma06: This won't work for a select, you have to range over each option tag and apply the selected attribute on the right one

PR #1191, opened 2017-07-14
Fix flag spacing that previously was done with a space instead of margin

Kiloutre:


akuma06:
Modify directly layouts/partials/helpers/flags.jet.html.

PR #1219, opened 2017-07-17
Automatically add flag based on keyword

Kiloutre:
screenshot
We get a pretty decent amount of french torrents and i'm fairly sure the "VOSTFR" keyword is unique enough to automatically put the french flag should the user add this keyword to his torrent
Same goes for [ITA], unique enough and gets enough uploads to make it into there
We could do this in Go, but i very much hate the idea of adding flags to a torrent without the user being aware of it. You can do that on scraped torrents if you want but don't on user-uploaded ones.
Removing the keywords does not remove the flags as it would be annoying afterward should an user copy paste a torrent name and modify it to his liking

### DIFF
--- a/public/js/kilo.js
+++ b/public/js/kilo.js
@@ -96,12 +96,25 @@ var Kilo = function (params) {
   var updatePreviewTorrentName = function (e) {
     var el = e.target
     self.setName(el.value)
+
+    var Keywords_flags= [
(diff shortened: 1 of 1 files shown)


akuma06:
kilo you should probably run the for in an anonymous function called by a setTimeout so it won't make the user hang on the keyup event

PR #1293, opened 2017-07-27
Removal of JS for searchbox & refine

Kiloutre:
I've found myself spending a lot of my time browsing the website with JS disabled recently, and here's why:

The searchbox JS, although designed to be pratical, ended up being more annoying than anything. Proceeding to write your text then clicking the search icon only to see the search box suddenly shrink and the form not submitting is a very annoying thing, and i can't really think of any viable workaround. Is the searchbox expanding really needed or even wanted? I can't think of any time it's been useful to me, and just like for other users it's been annoying a few times for me too.

As for the refine, i just don't really think the JS is needed nor useful. There's nothing inherently wrong with one page load to get to the refine stuff, and the fact that you can in the header select the category, write in the searchbox and then click refine to start the search while now showing the refine options (if JS disabled) makes much more sense to me: you get to see relevant results immediately and can look at them to start thinking about how precisely you are gonna refine it (which flags? trusted/not?), as it all depends on what content there already is without refining the search. I personally like the behavior of the refine stuff without JS disabled much more, and as stated in the first line it's pretty much why i disable the JS 100% of the time while browsing the website.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -123,16 +123,6 @@ function startupCode() {
 
   if (document.cookie.includes("newVersion"))
     document.getElementById("commit").className = "new";
-
-  document.getElementsByClassName("search-box")[0].addEventListener("focus", function (e) {
-    var w = document.getElementsByClassName("h-user")[0].offsetWidth
-    document.getElementsByClassName("h-user")[0].style.display = "none"
-    document.getElementsByClassName("search-box")[0].style.width = document.getElementsByClassName("search-box")[0].offsetWidth + w + "px"
-  })
-  document.getElementsByClassName("search-box")[0].addEventListener("blur", function (e) {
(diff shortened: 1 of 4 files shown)


akuma06:
It's a hassle to have to click again on refine and be moved on another page when you can actually limit that behaviour by a little js button "show refine". That's the one of the point of JS, making things faster. In fact it is inherently wrong to do one page load to get just to the refine div displayed. If you look how forums/search works on other website you will see that nobody sane does it like that. Forums would lead you to an "advanced search page with no results whatsoever" and the old nyaa would display by default the advanced search form whenever you made a search. 
If you add to that that websites would also make the same behaviour as nyaa but with a showable div in JS for js users 
Anyway I don't mind actually removing the refine button from the header (never liked it there). But I don't like removing the js behaviour for refining. It's like moving to a website behaviour from the 90's with not even the refine div shown by default

PR #1304, opened 2017-07-31
Initial Tag System [done]

akuma06:
This adds a basic support of tags with Pantsu Points.
* User have a starting number "1" of pantsu points
* Torrents have tags composed with a "tag name" and a "tag type" (eg. anidbid = "tag type" and 124521 = "tag name")
* So torrents can have multiple "tag name" of the same "tag type"
* Enabled Tag types are defined in config.yml 
* On view page, a user can add a tag by filling a form.
* On view page, a user can vote for or remove his vote for a tag.
* When voting for a tag, the system calculates how much votes have been collected for the tag. And if the sum has reached the maximum value set in config.yml, the other tags of the same type are removed, the users having voted wrongly have their pantsu points decreased and the others have their pantsu points increased.
* Tags on view page are differentiated between "accepted tags" and "votable tags"


Kiloutre:
> Pantsu points

the idea **sounds awful** 🐤

PR #1306, opened 2017-08-01
Hiding pending tag for anon

Kiloutre:
And some css color change that still will need some work

### DIFF
--- a/models/torrent.go
+++ b/models/torrent.go
@@ -1,447 +1,61 @@
 package models
 
 import (
-	"context"
 	"errors"
-	"fmt"
-	"html/template"
-	"path/filepath"
-	"reflect"
-	"strconv"
-	"strings"
-	"time"
-
-	elastic "gopkg.in/olivere/elastic.v5"
-
 	"net/http"
-	"net/url"
 
-	"github.com/NyaaPantsu/nyaa/config"
-	"github.com/NyaaPantsu/nyaa/utils/cache"
-	"github.com/NyaaPantsu/nyaa/utils/format"
-	"github.com/NyaaPantsu/nyaa/utils/log"
-	"github.com/NyaaPantsu/nyaa/utils/sanitize"
-	"github.com/bradfitz/slice"
 	"github.com/fatih/structs"
 )
 
-const (
-	// TorrentStatusNormal Int for Torrent status normal
-	TorrentStatusNormal = 1
-	// TorrentStatusRemake Int for Torrent status remake
-	TorrentStatusRemake = 2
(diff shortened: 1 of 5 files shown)


templates/site/torrents/view.jet.html
```diff
@@ -64,24 +64,24 @@ <h1 style="text-align: center;" class="torrent-hr">{{Torrent.Name}}</h1>
...
-	{{ if len(Torrent.Tags) != 0 || User.ID > 0 }}
+	 {{ if Torrent.Tags.HasAccepted() || User.ID > 0 }}
     <tr class="torrent-info-row">
       <td class="torrent-info-td torrent-info-label">{{ T("torrent_tags")}}:</td>
       <td class="tr-flag torrent-view-td torrent-info-data">
-        {{ range Torrent.Tags}}
+        {{ range Torrent.Tags }}
+		{{if !.Accepted || User.ID > 0 }}
         <span class="tag{{ if len(Torrent.Tags) > 5 }} big{{ end }}{{ if !.Accepted }} pending{{else}} accepted{{end}}" title="Tag: {{ .Type }} ({{ if !.Accepted }}{{ .Total }}{{else}}{{ T("accepted") }}{{end}})">
           <span class="tag-text{{ if User.ID > 0 }} votable{{end}}">
```
akuma06: Since the anon don't see votable tags, you don't need to add the condition

models/torrent.go
```diff
@@ -98,7 +98,7 @@ type TorrentJSON struct {
 	Completed    uint32        `json:"completed"`
 	LastScrape   time.Time     `json:"last_scrape"`
 	FileList     []FileJSON    `json:"file_list"`
-	Tags         []Tag         `json:"-"` // not needed in json to reduce db calls
+	Tags         Tag           `json:"-"` // not needed in json to reduce db calls
```
akuma06: Just so you remember about the missing s

models/torrent.go
```diff
@@ -63,7 +63,7 @@ type Torrent struct {
 	OldUploader string       `gorm:"-"` // ???????
 	OldComments []OldComment `gorm:"ForeignKey:torrent_id"`
 	Comments    []Comment    `gorm:"ForeignKey:torrent_id"`
-	Tags        []Tag        `gorm:"-"`
+	Tags        Tag          `gorm:"-"`
```
akuma06: here too for the missing s

templates/site/torrents/view.jet.html
Kiloutre: Yes i do, if i understood it right there may be both Accepted and Pending tags for the same torrent

Issue #1334, opened 2017-08-03
Search argument changes backward compatible in v1.1 + deprecated in v1.2


akuma06:
This issue is made to announce that the website will move from our current way of limiting search in time. Instead of doing "fromDate=xx" and "toDate=xx" with "dateType=xx" arguments, we will accept also "fromDate=YYYY/MM/DD" and "toDate=YYYY/MM/DD" in v1.1, and, in v1.2, remove the previous way of doing the search.

# What will actually change?
Before when doing a refined search, you had to type the number of days, months, years, you wanted the search to take in order to limit the search in time. For example, you had to actually do: "From: 9 To: 10 DateType: Days" to search for results from 9 days to 10 days ago. This behavior was highly criticized and said hard to use due to computational logic in it.
In final, the search arguments in the url where like that: "?q=xxx&fromDate=9&toDate=10&dateType=d".

People were asking to directly be able to input dates. Therefore, we will change progressively the behavior stated previously to dates input in v1.1.

# What does it mean?
For example, the form would display like this: "From: 2017/07/02 To: 2017/08/27". No more dateType field. This means you're looking for torrents uploaded between 2nd July 2017 and 27th August 2017. 

# I'm using the old urls, would that mean I won't be able to use that?
We will provide backward compatibility until v1.2 is pushed. We are now in v1.0.1, next version v1.1 introduces the new search arguments and v1.2 will remove the old search arguments.

# Great so should I move from it?
You should as soon as possible. The backward compatibility is a momentary one, we don't want to have a too complicated code in our search functions. So removing it from there is a high priority. We won't delay the removal further than v1.2. Please remove it from your rss feeds/clients and use the new one.

# But I still want relative time search!
You can use the maxage argument which happened to have the same behavior as old nyaa. This argument will filter the torrents to the ones uploaded the last x days.

PR #1342, opened 2017-08-04
Tag Search + Tests + Search slight refactor [DONE]

akuma06:
First commit improving search. Different struct have their own file with their tests. This way of separating struct by files is inspired by the go packages I've seen so far.

Added new behaviour as discussed in #1334
Now there will be descriptive tags for a particular release (ecchi, BDSM, ....) and typed tags.
Typed tags are tags relevant to all torrents and can be limited to some input value. For example, video quality is a typed tag limited to some values (hd, full hd, sd, ...). In the same way, anidbid is also a typed tag but doesn't have default values.

Furthermore, the location storage of tags have changed, now accepted descriptive tags are stored in the torrents table in the column "tags" and they are separated by commas.
In the opposite, accepted typed tags have have their own column in the torrents table. For example, anidbid, vndbid will populate the column AnidbID, VndbID when accepted.

This behaviour depends on the callbackOnType function in tag/helpers.go


ewhal:
@akuma06 what needs doing for this to be finished?

akuma06:
Not much I have some things on my computer to push. But I'm away from home so I can't push them :'). Will be back tomorrow 👍

PR #1345, opened 2017-08-04
Translation Overhaul #2

Lamppostmanufacturer:
This concerns these other translations that were not upto-date: it-it, nl-nl, pt-br, pt-pt, ro-ro, th-th

Note: No actual Translations were made.

These translations are now on par with the en-us one. Old strings weren't overwritten.
Its now a shitton easier for the translators of their respective country to translate the strings now since all the mess of old unorganized strings is properly aligned to the english translation for better future translations.

### DIFF
--- a/translations/it-it.all.json
+++ b/translations/it-it.all.json
@@ -1,5 +1,25 @@
 [
   {
+    "id": "rules",
+    "translation": "Rules"
+  },
+  {
+    "id": "no_cp",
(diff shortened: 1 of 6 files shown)


Lamppostmanufacturer:
Its just a feature. Some of those translations were missing 1000+ lines and some changes are in the middle of other strings. Totally messed ordererd etc. This is to ensure all translations are on par with each other and since empty ID's will revert to the original file anyway it makes no difference other than easier management.

PR #1344, opened 2017-08-04
add Indonesian translation

39-chan:
I've translated the translation file (from English) into my language

[in-id.all.zip](https://github.com/NyaaPantsu/nyaa/files/1199915/in-id.all.zip)

PR #1350, opened 2017-08-04
Add Indonesian translation, credits go to @39-chan

Lamppostmanufacturer:
Thanks @39-chan for helping us out


akuma06:
Close #1344 

Only need an indonesian flag before merging.

PR #1356, opened 2017-08-05
Add RSS Feed Information to FAQ

yiiTT:
Also add translation support? But it should be self-explanatory to anyone reading it.


akuma06 (dismissed):
Need to add translation strings :')

jithatsonei (approved):
make them templated in a later PR pls

jithatsonei (dismissed):
i lied template now you cuck

ewhal (approved):

PR #1357, opened 2017-08-05
Add indonesian flag

Kiloutre:
For #1350 
flagmaking


public/css/flags/flags.css
```diff
@@ -93,10 +93,14 @@
     background-position: -80px -11px;
 }
 
-.flag.flag-multiple {
+.flag.flag-id {
     background-position: -80px -22px;
 }
 
+.flag.flag-multiple {
+    background-position: -80px -33px;
```
jithatsonei, line 101: should be precentage rather than a static number ideally ( for different resolutions) but ill let it slide

jithatsonei (approved):

public/css/flags/flags.css
akuma06: Set two css variable for the height and width of a flag. Then you can also use them to determine which flag to show. :p

PR #1371, opened 2017-08-14
Turn SearchType variable into an int

Kiloutre:
not that it matters much but heh


controllers/search/search.go
```diff
@@ -64,6 +64,10 @@ func SearchHandler(c *gin.Context) {
 	if c.Query("order") == "true" {
 		searchForm.SortOrder = true
 	}
+	searchForm.SortType, err = strconv.Atoi(c.Query("sort"))
```
akuma06, line 67: Move this to https://github.com/NyaaPantsu/nyaa/pull/1371/files#diff-2ff0b4807eadb9a31e1e7b187f5da802R50
It's better to have it directly set in the NewSearchForm function

akuma06 (changes requested):

controllers/search/search.go
Kiloutre: order & sort are not related to the search results nor form but only to the refine form, more precisely only to show the correct value in the two \<select\> after a page load which is just a display thing, so putting it right under the refine bool which is also just a display thing (display refine form yes/no) seems to make sense to me

akuma06: It's not that, NewSearchForm have an argument with c *gin.Context. This function should provide a default SearchForm based on the values passed by c.
Except for refine which I agree should be set in SearchHandler, I would like order&sortType to be set in NewSearchForm. The reason is that I would like to reuse the SearchForm for things beside the Search pages (for example, in mod panel for listing pages).
Therefore, in order to not copy/paste multiple times the lines for "order" and "SortType", having them directly in NewSearchForm is better.

Kiloutre: alright but i already tried making the changes you requested and i miserably failed at it so i can't really do it anyway

PR #1381, opened 2017-08-20
Add toggle dark theme for non-JS users

Kiloutre:
Having it only for JS users is pretty shit, after all we're not like (((them)))
If user has JS disabled, the toggle will now redirect to /dark page which will toggle the dark theme then redirect to the page the user was when clicking on the toggle


akuma06 (approved):
Looks good

PR #1393, opened 2017-08-22
Removing filter on listing.jet.html

akuma06:
Filtering here would mean giving random number of results instead of showing constant number. And can mislead people to think that there are no results and also force people to switch pages more often.
For example, if there are 48 hentai results on the 50 results. It will display only 2 results on the page. People would therefore think that there are only two results for the search or would make them changing the page every two (or more) results instead of 50.
Finally this doesn't fix at all the results in the RSS feeds & the API endpoint.

The hentai results should be fixed in the scrapper (done).

### DIFF
--- a/templates/site/torrents/listing.jet.html
+++ b/templates/site/torrents/listing.jet.html
@@ -52,7 +52,6 @@
   </thead>
   <tbody id="torrentListResults">
     {{ range Models}}
-	{{ if CategoryName(.Category, .SubCategory) != ""}}
     <tr id="torrent_{{ .ID }}" class="torrent-info {{if .Status == 2}}remake{{else if .Status == 3}}trusted{{else if .Status == 4}}aplus{{end}}" >
       {{ if User.HasAdmin() }}
       <td class="tr-cb hide">
@@ -106,7 +105,6 @@
         {{end}}
         <td class="tr-date home-td date-short hide-xs">{{.Date}}</td>
       </tr>
-	  {{end}}
       {{end}}
     </tbody>
   </table>


Kiloutre:
I disagree with this.
https://nyaa.pantsu.cat/search?c=2_&userID=0&q=
pic
Browsing the audio categories is pure 100% hell with these garbage torrents around
Yeah, it's not a fix but a workaround and is certainly not ideal, but it's much much better than the shit we have on live right now
I browse the audio category frequently and i'd rather see only 5 torrents per page than having to scroll so much because of garbage, completely unrelated torrent
Showing only 5 good torrents is objectively much better than showing 5 good and 45 completely unrelated and uninteresting ones
The scrapper problem isn't going away anytime too, and it's idealist to think it will.
If you want an actual fix you prevent torrent with unmatching categories to pop up in the search period (instead of just hiding it as we are doing now), but deleting the filtering without submitting any fix yourself is pure idiocy.
The picture above has 7 normal torrents, followed by 40 garbage ones in a row and then 3 normal ones. Browsing this shit is purely inacceptable, and as inconvenient as showing only 10 torrents is it's still much better than showing extra useless ones.
after
This is already much better to browse, i can move on to the next page right away. The only things the garbage torrent bring are extra scroll time and more time spent looking at the torrents to identify the garbage and non-garbage ones.
I don't understand why you want this.

akuma06:
Like I said, showing 7 results on 50 because there are garbage results mixed in is not a **fix**, it's bad workaround and lead to wrong coding behaviours/other bugs. You don't introduce fix by introducing new bugs.
And it actually fix nothing since the results in rss&api search are not impacted by this. So the results from them and the results by a website search are not consistent.
Finally, by doing this you lead to randomized number of results and force people who wants for example to search between 300 results to look through much less (even if there are garbages mixed in). 

If you want a fix, fix scrapper (which is fixed according to @ewhal).

Kiloutre:
My man the picture i posted earlier show 40 garbage torrents that should not be shown in there out of 50, do you really think the user gives a shit if there's less than 50 torrents as opposed to 4/5 of them being 100% useless?
Start thinking more like an user, there is nothing more irritating than browsing the audio categories only to find it filled with non-audio torrents. Yeah, it's not showing the 50 or more torrent it should show, what about it? Do you seriously think showing garbage torrents is better than that? Do you really think swimming through garbage torrents just to find new content is okay whatsoever? Not showing these useless torrent is infinitely better for the user, it removes one massive inconvenience to introduce a very small one.

More importantly, you obviously have the knowledge to fix this yourself. Counting on ewhal to fix this is stupid: not only does he have trouble fixing it every time it happens, but a simple a change on anidex's html would completely break the scrapper again. You could make a PR right now that would simply ignore garbage torrents during search in order to show the right amount of torrents, but you're not doing it. This problem isn't going away and removing a workaround without fixing the problem is useless.

Showing garbage torrent > Showing a random number of torrent per page, inconvenience wise. From a dev point of view it is indeed messy but from an user point of view it is more convenient.

akuma06:
Like I said, ewhal said it's already fixed so why are you complaining?
For anidex html changes, the scrapper can easily filter out hentai when parsing the result by having a cookie: "anidex_h_toggle=1" which is how anidex knows not to show hentai torrents.
Moreover, the scrapper adds the hentai torrents in a non-existant category on nyaa, another easy fix for this was to just check the category before adding it.
So a change of the anidex html occuring another bug like that could easily be prevented by some checks and more profound filtering. This has already in part been implemented in the scrapper by ewhal.

Furthermore, if you want to easy fix this, just remove the torrents with category=2 and sub_category=2 in the database. This has already been given by sfan5 as a fix for these hentai torrents.
Doing the filter on the fly is however not a fix (especially when you do it only for the torrents listing and not for the RSS & API search). Fixing scrapper (done) and removing hentai from database (can be easily done by a simple query, `DELETE FROM torrents WHERE category=2 AND sub_category=2` ) are better options.
Making a filter to prevent category from appearing on the search function would make it more complicated that it is already and other solution would use more cpu ressources. When you have solutions that can ease the burden of the program to achieve the same results, you choose them. 
And finally, not showing garbages torrents introduce another thing, moderators won't know that there is garbages torrents mixed in. And don't tell me "if there are not enough results they can think that there are garbages torrent", because not showing enough results can lead to many more problems. For example, a bug in Elastic Search, a corrupted database, a function in Golang that is broken due to some unexpected data handling. All this leading to more difficult bug resolution.

And don't give me the "think of a user perspective", if you think from this, you should know many uses RSS to actually download torrents automatically with their torrent client. This doesn't fix it.

akuma06:
I forgot that moderators can remove the torrents easily also with the modtool bar. This is also another way to prove that knowing and showing when there are garbages is important in order to moderate them (if it still happens in the future).

Kiloutre:
>Like I said, ewhal said it's already fixed so why are you complaining?

Because the garbage torrents are still there?
You'd know if you actually used the website ;)
And regardless it's not like this issue is never gonna pop up again so removing the workaround without submitting a fix yourself is silly.

>Furthermore, if you want to easy fix this, just remove the torrents with category=2 and sub_category=2 in the database. This has already been given by sfan5 as a fix for these hentai torrents.

It doesn't happen only with the audio (2) category, you'd know if you actually used the website ;)

>And don't give me the "think of a user perspective",

too late ;)

>For example, a bug in Elastic Search, a corrupted database, a function in Golang that is broken due to some unexpected data handling. All this leading to more difficult bug resolution.

The only thing the workaround does is not feed the torrent HTML should it be invalid, how in the fuck could that cause any of the thing you just quoted?
It's not like it's using some extremely fancy function to check if the torrent is garbage, it uses the CategoryName() function that will get called anyway for the title of the category image, there is no possible way to have an unexpected data handling in there or a Golang function breaking or any other silly thing you may think.

akuma06:
"It doesn't happen only with the audio (2) category, you'd know if you actually used the website ;)"
As long as the category is not one of nyaa, you just have to make the delete query with the category & sub_category set.
And if it happens with a category set in nyaa, then your code won't remove the torrent, since the category exist.
Of course they are still there but they shouldn't be there in the future. Just removing those already in database is easy. And checking before adding the torrents in database that the category you're actually adding is in nyaa fix the problem (and scrapper is fixed). Even a change in anidex html or anything else can't produce the bug.

When I spoke about ES bug and everything else, I'm speaking about when you resolve bug. Having the wrong number of results can make devs believe that what they are doing causes display bugs when they test their code. Or on the contrary, a bug that have happened can make people believe that there are no bugs and it is just the code that filter the thing. Anyway even if the possibility is not high, I don't see why you would still have this when you don't need it anymore. Because it is fixed and if, by chance new hentais does appear on nyaa, they would be added on already existing categories which would be not prevented by this code.

ewhal (approved):

PR #1394, opened 2017-08-22
Add tag inputs to refine search

Kiloutre:
owo


akuma06:
@Kiloutre for quality can you add a select instead?

Kiloutre:
@akuma06 
can't because i don't know the options the select should have

ewhal (approved):

PR #1398, opened 2017-08-25
Fix Navigation Admin Panel

akuma06:


Kiloutre:
`title="{{ .User.Username }}"`

What is the point of an element having the same title as it's own content? I don't think there's much point in showing a "kilo" title on a "kilo" text

Kiloutre:
Add  `{{ if Form.Magnet != "" }}class="hidden"{{end}}`
to `<h3>{{ T("torrent_file")}}</h3>`

ewhal (approved):

PR #1399, opened 2017-08-25
Exclude User Search argument

### LINKED ISSUE #528: Exclude torrents in search by user id.
On the old site there was an input field to hide torrents by user id, this is sadly missing from the new site. This was one of the features I used the most because it helped remove all the clutter of all the HorribleSubs releases.

akuma06:
Now you can exclude a particular user from search by putting in the url ?nuserID=xxxx
Closes #528


Kiloutre:
What for??
Could you add search by username?

akuma06:
It is a very old issue that reproduce old nyaa behavior. Just read the issue :')

ewhal (approved):

PR #1404, opened 2017-08-26
fix inline CSS not working

Kiloutre:
thanks akuma


akuma06:
Don't thank me, I requested a review on it and not my CSP code. Only got it from the issue linked to my PR :')

akuma06 (approved):

ewhal (commented):
You'll also want to modify img-src and media-src iirc

PR #1405, opened 2017-08-26
Fix refine date inputs being too small

Kiloutre:
Didnt have enough room for the text it was trying to contain

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -425,6 +425,9 @@ select.form-input {
     width: 47%;
     position: relative;
 }
+.refine-date {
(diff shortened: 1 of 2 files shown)


akuma06:
Isn't that because size=7?

Kiloutre:
> Isn't that because size=7?

Yes, but that itself depends on resolution, e.g size=8 will look good on 1080p but way too small on very low res. But yeah i completely overlooked it, i'll remove it although it will not change anything

ewhal (approved):

PR #1407, opened 2017-08-27
fix base64 encoded imgs

ewhal:


yiiTT (approved):
Matches CSP spec for base64 images

PR #1440, opened 2017-08-29
"Alternating Colors" option in settings

Kiloutre:
Note: **Disabled by default**
Note2: **Travis is failing but the code works, just run it locally**
I recall getting quite a lot of user complaints in the past during the time of the wide header crisis, in which people called out how retarded it was to use bold text on every single torrent name as it made the text much longer than necessary, so i thought about the option but after a quick test it turned out pretty annoying to look at a torrent list without bolded fonts and with all the torrents having the same background (really hard to keep track of anything), at which point i realised that having alternating colors would fix this problem and as such would allow the font to be of normal size instead of bold
I made it into an user settings, just like any user settings it works for both logged in and non-logged-in users
It might need some work when it comes to the colors of trusted, remake and a+ torrents, and also for the tomorrow colors, but it's already a good start in my opinion
Screenshot colors are out of date 

g css
tomorrow
classic
Here is screenshot of the select in the settings:
settings


Lamppostmanufacturer:
Thanks for more customization settings. I personally would align the changes stuff like we did on the advanced search instead of stacking it up. Maybe in 2-3 columns? Also considering there will be more and more settings to make it would make sense to have them all on a static page rather than one where you'd have to scroll.

ewhal (approved):
bretty gud/10

PR #1451, opened 2017-08-30
Mascot always clickable

emanueln:
Changed three z-index style rules so mascot would always be clickable, as per bug reported on May 29

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -37,7 +37,7 @@ body {
 
 #mascot {
     position: fixed;
-    z-index: 1;
+    z-index: 3;
     bottom: 0;
     right: 0;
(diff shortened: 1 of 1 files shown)


public/css/main.css
```diff
@@ -1867,6 +1867,9 @@ span.tag {
 html[lang="ja-jp"] .header .nav-btn { padding: 0 8px;}
 @media (max-width: 1100px) {
     html[lang="ja-jp"] .header .h-user>.nav-btn {
-	font-size: 75%;
+    font-size: 75%;
+    }
+}
+e: 75%;
```
Kiloutre, line 1873: copy paste mistake here (very bottom of the file)

PR #1462, opened 2017-09-02
Fix broken sub category search

### LINKED ISSUE #1428: [dev.pantsu only] 500 Internal Server Error when searching for trusted/a+ or remake
https://dev.pantsu.cat/search?s=2
https://dev.pantsu.cat/search?s=3
https://dev.pantsu.cat/search?s=4

When searching by trusted, a+ or filtering remake, an error 500 happen on dev. It does not happen locally though, even when both local & dev are using the exact same code

akuma06:
Fix #1428


Kiloutre:
groundzeroes

Kiloutre (approved):

Kiloutre:
btw this doesn't FULLY fix #1428 as the issue also mentions an error 500 when searching by trusted, remake or a+

PR #1468, opened 2017-09-02
Fix adding tag form (nojs) + miscalleneous

### LINKED ISSUE #1423: Constant error: The tag type selected doesn't exist on /torrent/tag
https://dev.pantsu.cat/torrent/tag?id=955378
The form also straight up doesn't add the tags, even if we ignore the issue with the wrong value in the select options (still happen if value = "anidbid")

akuma06:
* Fix #1423 by checking http method before finding the tags.
* Form is now like other tags form
* New Middleware LoggedInMiddleware to protect routes to logged user only
* Display of Non-Accepted tags in torrent view improved by translating tagtype values


Kiloutre (approved):
Works well, takumi!

PR #1499, opened 2017-09-04
Add background color to remake-checkbox

MMP0:


Kiloutre:
I'm not against adding a background for the remake checkbox that would be colored like remake torrents, however it'd need more designing than simply applying a class="remake"
remake
remake2
Also, the **for="remake"** needs to stay, it's there so that you can click on the text in order to check or uncheck the checkbox and as such is pretty useful

ewhal (approved):

PR #1504, opened 2017-09-05
Add torrentFileExists function, check if torrent file exists and if it doesn't lower the opacity on the torrent download button

Kiloutre:
Checking if TorrentLink is empty is useless because it never is, even if the file doesn't exist
Also i like visual cue that show it should exist but doesn't

@akuma06 i'm encountering problems with making the function available in listing & view.html :(
pls help


templates/template_functions_test.go
```diff
@@ -588,6 +588,28 @@ func TestContains(t *testing.T) {
...
 }
 
+func testTorrentFileExists(t *testing.T) {
+	var tests = []struct {
+		hash 	     string
+		Expected     bool
+	}{
+		{
+			hash: "98C5B0FC5FC996A565D5DFE2312C467213BCC2EB",
+			Expected: true,
```
akuma06: You can't test "true" since you don't provide the torrent with hash "98C5B0FC5FC996A565D5DFE2312C467213BCC2EB" in the source code, so the torrent doesn't exist.
You should test "" return false. And I think it's enough.

akuma06 (changes requested):

templates/template_functions_test.go
```diff
@@ -588,6 +588,28 @@ func TestContains(t *testing.T) {
...
+		},
+		{
+			hash: "38C5B0FS5FC996B565D5EFE7312C4673BCX9MS",
+			Expected: false,
+		},
+	}
+	for _, test := range tests {
+		value := torrentFileExists(test.hash)
+		if value != test.Expected {
+			t.Errorf("Unexpected value from the function TorrentFileExists, got '%s', wanted '%s' for '%d'", value, test.hash, test.Expected)
```
akuma06: Also change to:
```
t.Errorf("Unexpected value from the function TorrentFileExists, got  '%t', wanted '%t' for '%s'", value, test.Expected, test.hash)
```

akuma06 (changes requested):

templates/template_functions.go
```diff
@@ -292,3 +294,13 @@ func contains(arr interface{}, comp string) bool {
 	}
 	return false
 }
+
```
akuma06, line 297: In this file, line 35 you need to set Vars: `vars.Set("torrentFileExists", torrentFileExists)`

akuma06 (changes requested):

templates/template_functions.go
Kiloutre: totally forgot vars.set also needed to be done for functions and not just variables

templates/template_functions.go
```diff
@@ -32,7 +32,7 @@ func templateFunctions(vars jet.VarMap) jet.VarMap {
 	vars.Set("getDefaultLanguage", publicSettings.GetDefaultLanguage)
 	vars.Set("FlagCode", flagCode)
 	vars.Set("getAvatar", getAvatar)
-
+	vars.set("torrentFileExists", torrentFileExists)
```
akuma06: Uppercase S: `vars.Set`

akuma06 (dismissed):

akuma06 (approved):

PR #1506, opened 2017-09-05
Show message pinned at the bottom of the window warning of cookie usage to comply with EU law

Kiloutre:
No preview because i haven't designed anything yet, will do once it gets merged


templates/template_test.go
```diff
@@ -288,6 +288,7 @@ func mockupCommonvariables(t *testing.T) jet.VarMap {
 	variables.Set("User", &models.User{})
 	variables.Set("URL", &url.URL{})
 	variables.Set("CsrfToken", "xxxxxx")
+	variables.Set("EUCookieLaw", "test")
```
akuma06: Set this to a boolean

utils/publicSettings/publicSettings.go
```diff
@@ -237,6 +239,18 @@ func GetMascotURLFromRequest(c *gin.Context) string {
 	return ""
 }
 
+func GetEUCookieFromRequest(c *gin.Context) bool {
+	cookie, err := c.Cookie("EU_Cookie")
```
akuma06: Since you don't use cookie, you just have to do:
_, err := c.Cookie("EU_Cookie")

akuma06 (changes requested):

templates/template_test.go
```diff
@@ -288,7 +288,7 @@ func mockupCommonvariables(t *testing.T) jet.VarMap {
 	variables.Set("User", &models.User{})
 	variables.Set("URL", &url.URL{})
 	variables.Set("CsrfToken", "xxxxxx")
-	variables.Set("EUCookieLaw", "test")
+	variables.Set("EUCookieLaw", "false")
```
akuma06: No I mean, don't use "false", but `false`

Kiloutre: i know it was just a mistake

akuma06 (approved):

yiiTT:
Kilo you really like to remove coverage

PR #1508, opened 2017-09-05
Update tracker and Ad's explanation.

Miiiiaaooooo:


Kiloutre (dismissed):

translations/en-us.all.json
```diff
@@ -1893,7 +1893,9 @@
   },
   {
     "id": "faq_ads_explanation",
-    "translation": "[lengthy explanation as to why we have ads"
+    "translation": "This is not a decision we make lightly, and we have thoroughly considered the objections raised by all of you, but ultimately we feel this is the only sustainable solution to fund Pantsu and ensure its continued operation. \
```
akuma06: Use `<br>` html tag instead. JSON format doesn't allow line-breaks.

akuma06 (changes requested):

translations/en-us.all.json
Kiloutre: I'm still keeping the \ because having an extremely long line isn't really good for translation purposes

akuma06: In JSON format, you can't have line-breaks in string. You have to put every string in one line (only). That's how the format works. And that's why your travis fails

akuma06 (dismissed):

translations/en-us.all.json
akuma06: https://stackoverflow.com/questions/2392766/multiline-strings-in-json

Kiloutre (approved):

translations/en-us.all.json
yiiTT: doko.moe > yw0.xyz has already been done across all translations FYI

yiiTT:
doko.moe > yw0.xyz has already been done across all translations FYI

PR #1568, opened 2017-09-15
Fix visual artifact on pagination when it was empty

Kiloutre:
It had some padding so was visible even when empty, aka showing an empty half-box


sfan5 (approved):
meme

PR #1571, opened 2017-09-16
Sync JS cookies & Go cookies's domain & random other things

Kiloutre:
We want to avoid same-name cookies using different domains because that creates annoying conflicts, preventing theme Toggler & user settings to work properly

- Fix cookies being all over the place when multiple of same name but different domain & value exists
- Fix header RSS URL
- ~~Fix \<br/\> issue in torrent description~~
- Fix too big spacing between ad & header, consistent with spacing between ad & main \<div\>
- Don't show top ad if there's an announcement, it's way too cluttered
- Fix "unknown" stats that were shown as "0 0 0"
- Fix cookie warning never showing up
- ~~Workaround for DLSite link not working, works for RJ links~~

### DIFF
--- a/models/tag/helpers.go
+++ b/models/tag/helpers.go
@@ -3,6 +3,7 @@ package tags
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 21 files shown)


yiiTT (dismissed):

yiiTT (dismissed):

yiiTT (approved):
wide scope PR approved
"yes it works" -kilo

Issue #1517, opened 2017-09-06
On demand torrent file generation for missing torrent files


ewhal:
https://github.com/anacrolix/torrent/blob/master/cmd/magnet-metainfo/main.go

Kiloutre:
@ewhal plenty of the new users complaining about the torrent links not working
you're the only one that has enough knowledge to do this
pls

Kiloutre:
Also you could directly run the script in download.go should the file be missing, that way we could straight up avoid having it run automatically on every torrent but rather generate files for torrents that actually need it

PR #1600, opened 2017-09-19
Torrent Generation on not found error

akuma06:
As asked in #1517, it allows on-the-fly torrent generation. Since it uses magnet links, it needs some time to connect to peers. So it can't be instant generation, we need the user to wait and try after a minute at least.

Edit: ~~Seems like I can't test the PR on my windows machine because Anacrolyx torrent doesn't want to compile...~~
Edit 2 : Works on windows if you have gcc installed, I used msys2 and its tutorial for that https://github.com/orlp/dev-on-windows/wiki/Installing-GCC--&-MSYS2


Kiloutre:
I have no idea what ewhal exactly did, he didn't even commit the changes here, but torrent file generation is already there and working for seemingly every single torrent
Not sure if done on download.go or if he has a script running somewhere that does it for every one of them though
But dont work on this any more until @ewhal gives clarifications as it's a potential waste of time
(Hence why i closed the issue)

ewhal:
The torrent generation script is shit so this is a very good idea

akuma06:
Si quelqu'un peut tester please =)

Kiloutre:
I tried already yesterday but i get a GCC error when go building

akuma06:
Yeah it seems anacrolyx doesn't compile on windows. We need someone with linux :/

Kiloutre (dismissed):
haven't tested but in case you wanna merge it still

akuma06:
I don't think we should, it will prevent people to build nyaa on windows :/

akuma06:
Ok to build on windows you need gcc. I just successfully build with the help of msys2 and this tutorial :  https://github.com/orlp/dev-on-windows/wiki/Installing-GCC--&-MSYS2

akuma06:
Tested on my end and it works

controllers/torrent/download.go
```diff
@@ -7,27 +7,41 @@ import (
...
 	"github.com/NyaaPantsu/nyaa/utils/format"
 	"github.com/gin-gonic/gin"
 )
 
 // DownloadTorrent : Controller for downloading a torrent
 func DownloadTorrent(c *gin.Context) {
 	hash := c.Param("hash")
+	messages := msg.GetMessages(c)
+
+	if hash == "" { // if no hash provided, you can't find a torrent in db neither the torrent file
```
Kiloutre: There is no /download/ route, only /download/:hash
If hash is empty then download.go will never get executed in the first place, i think?

Kiloutre:
>Works on windows if you have gcc installed, 

Golang/git already requires GCC to be installed on windows, your branch though still gives me the error

akuma06:
>>Works on windows if you have gcc installed,
>
>Golang/git already requires GCC to be installed on windows, your branch though still gives me the error

Anacrolyx requires to have the last gcc and mingw. That's what I meant. The only one that gives me a compatible version on windows is through Msys2. So if you want to compile this branch, move to Msys2 or a gcc version similar to the one provided through Msys2.

controllers/torrent/download.go
akuma06: You're right but I don't know if we should really not check it.

Kiloutre (dismissed):
i tried but couldn't manage to make it work, i kept getting errors :(

akuma06:
@Kiloutre, you should remove every mingw/gcc from your computer (remove TDM if you used that or Mingw or Cygwin). After that check that in CMD, when you put the command "gcc", it says command not found. Now install Msys2 and do the tutorial I linked to.

Finally, change your path environment variable and add "C:\msys64\mingw64\bin" AND "C:\msys64\mingw32\bin" in this order.

To check that it works, launch a new CMD (the previous one will not work because it doesn't know the change in Path environment) and type "gcc". It should no more say to you "command not found"

controllers/torrent/download.go
Kiloutre: checked it

Kiloutre:
I got rid of the errors
However i'm not sure how the hwole thing works. At first it used anicache which obviously gave a 404, then i removed anicache from cache_link which made the TorrentLink be completely empty, and i got a warning on download.go that we did not store files
Then i put the localhost /download/:hash link as storage_link and it did not generate them still, regardless of whether cache_link was empty or not

This isnt an improvement over the current way of doing it, since the current way is just going through every torrent and generating the files which means torrent files are there before you even need them instead of having to wait

akuma06:
You have to remove the cache link from your config and add the storage link. This is how the config.yml is set on the production website.
So this is what you need to have in your config.yml : 
```
torrents:
  filestorage: G:\xxx\nyaa\downloads
  storage_link: http://127.0.0.1:9999/download/%s
  cache_link:
```
filestorage should be your real path where you want all the .torrents
storage_link is the url of the download

Kiloutre (dismissed):
Alright it works
What do you think of the following JS things we could do before merging:
- ~~Click on the torrent link, instead of going to the torrent missing page we do a query on the download page, add a loading image in the download button and start the download once it is ready~~ (done)
- Same as above but triggered automatically

Kiloutre:
If a torrent file is generated then manually deleted, further attempts at generating it will not do anything and only restarting the server can fix it

Kiloutre (dismissed):
I think we are good
By the way at very rare times i get a complete fuckup when the script tries to generate a torrent, aka the whole server going down
It happened twice to me for a total of 30 generation i think, no idea what triggers it
It leaves me pretty worried if it's not only on my end

Kiloutre:
Do we really want to have it on anicache by default? I think it should be like this by default
```
torrents:
  filestorage: ./downloads
  storage_link: /download/%s
  cache_link:
```
That'd avoid having to configure it for it to work, it'd work right away

Kiloutre (dismissed):

Kiloutre (dismissed):

Kiloutre (dismissed):

Kiloutre:
Alright i'm getting problems with the script, i can generate one torrent file but after that it doesn't generate any more

akuma06:
It might be related to windows emulation, it seems the error is windows coroutines related. It would be great to test it on linux :). The easy way to make the error happen is to launch a torrent download and refresh just after before the torrent has been generated.

Kiloutre:
Even if it's windows-specific, pantsu is supposed to be able to be ran on anyone's machine, so it would be nice to get it working on windows as well
~~Besides, i think it worked fine a bunch of commits ago? I'll download an old one and see if it works~~

Kiloutre:
Ran it on dev and it does not seem to work either

akuma06:
Ok will take a look at it when I have time =)

ewhal (dismissed):

Kiloutre:
no
not working
no merging

Kiloutre:
Seems to work, gonna try on linux now

akuma06:
@Kiloutre ??

Kiloutre (approved):
Sry i needed ewhal to test and he simply he isn't there so i couldn't do it
gonna approve anyway so that you can merge if you want

Issue #1599, opened 2017-09-19
Allow for uploading to anidex, TT and other sites on upload


ewhal:
Similar to how anidex does it for TT
@akuma06

Kiloutre:
#1633

PR #1604, opened 2017-09-20
[WIP] Anidex, TT & Nyaa.si uploading

Kiloutre:
#1599 
@akuma06 @ewhal 
pic

- User goes in upload, adds his anidex API key and/or his nyaa.si API key to the inputs with the names 'anidex_api' and 'nyaasi_api' **OR** checks the corresponding checkboxes allowing him to upload as anon, submits form
- Torrent successfully uploaded to pantsu

Two ways we can go from here:

1. Queries to nyaa.si & anidex done in controllers, then the upload_multiple.jet.html page is shown with the status of each torrent. If the upload has failed, give error message and possibly a retry upload button (that works both with and without JS)
2. upload_multiple.jet.html page loaded immediately, queries done entirely through JS, allowing fancy live-updating of status (Sending upload query... /  Pending / Retry) instead of just Finished / Error
3. `1.` but we detect if JS is enabled and if it is, we end up doing `2.`

Either way this page has to be shown at some point whether it is after upload or before

Also, here is upload form:
owo
Also here is new user edit form
user edit

Changelog:
- User edit form redesign
- Add two inputs on upload
- Remove "unknown" message on unknown stats, replace by three **-** , as the unknown's colspan was breaking mobile view
- Remove unnecessary class and tag in comment td
- NyaasiAPIToken & AnidexAPIToken  user variable
- IE: Fix buttons on user profile, fix torrent flags on view
- Don't load pagination div if empty
- Remove total uploaded size
- Remove useless template
- Create upload_multiple template


sfan5 (dismissed):
sugoi

sfan5 (dismissed):
sugoi desu

sfan5 (approved):

Issue #1620, opened 2017-09-28
On-demand stat fetching


Kiloutre:
1. User browse torrents, click on a torrent with Unknown stats
2. Website shows view.html page, all three stats are on Unknown
3. A query is done through JS to a stats?torrentid=XXX page, this very page will try to fetch for the latest stats instead of relying on DB or such
4. The query returns the actual stats and update the stats field in view.html, on top of the stats in the DB so that future search about the torrent will show the now fetched stats

PR #1621, opened 2017-09-28
On-demand stats fetching

Kiloutre:
- Only done if stats are unknown, on view.jet.html
- Done through JS as to avoid slowing down the page load
#1620


Kiloutre:
@ewhal @akuma06 how do i update the scrape table with new scraped data
just need to do that and the PR can get merged

controllers/torrent/stats.go
```diff
@@ -0,0 +1,71 @@
...
+			torrent.Scrape = &models.Scrape{uint(id), uint32(stats.Seeders), uint32(stats.Leechers), uint32(stats.Completed), time.Now()}
+			torrent.Scrape.Update(false)
+		}
+	}
+	
+	return
+}
+
+func UnescapeString(s string) string {
+	//Special characters are escaped using their hexa code and i have no idea what function unescapes this so i replace the characters
```
sfan5: ~~https://godoc.org/html#UnescapeString~~

templates/site/torrents/view.jet.html
```diff
@@ -271,6 +271,34 @@ <h1 style="text-align: center;" class="torrent-hr">{{Torrent.Name}}</h1>
 {{end}}
 {{ block footer_js()}}
 <script type="text/javascript" src="{{ URL.Parse("/js/modal.js") }}"></script>
+{{if Torrent.LastScrape.IsZero || formatDateRFC(Torrent.LastScrape) == "0001-01-01T00:00:00Z"}}
```
sfan5, line 274: both conditions are identical
rtfm https://godoc.org/time#Time.IsZero

sfan5 (dismissed):

controllers/torrent/stats.go
Kiloutre: it's not html escaping.

templates/site/torrents/view.jet.html
Kiloutre: Torrent.LastScrape.IsZero constantly returns false, i even made an issue about it a month ago

controllers/torrent/stats.go
sfan5: https://godoc.org/net/url#QueryUnescape

templates/site/torrents/view.jet.html
sfan5: then use just the second condition? idk

Kiloutre: i plan to fix .IsZero but as of right now we're doing it that way, that double condition exists multiple time in the file already, doing it the same way everytime is better because it allows to fix everything at once
consistency

sfan5: ¯\\\_(ツ)_/¯

sfan5 (dismissed):

sfan5 (approved):

PR #1631, opened 2017-10-03
Lower amount of torrents shown in user profile from 20 to 15, user not found page

Kiloutre:
- Add "User not found" error page
- Possibility of searching an user profile for it (write down username, get to user page)
- Writing an username instead of an userid in userprofile (/user/xxx/yyy) URL will no longer return the Anonymous profile, but rather get you to the user not found page
- Move themes in their own folder
- Fix ad showing scrolling bars at times
- Reduce maximum amount of torrent shown in user profile from 20 to 15


Kiloutre:
user not found
user profile

sfan5 (dismissed):

sfan5 (dismissed):

sfan5 (approved):

PR #1633, opened 2017-10-04
Upload to nyaa.si, anidex & TokyoTosho [done]

Kiloutre:
#1599


config/structs.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 	I18n I18nConfig `json:"i18n" yaml:"i18n,flow,omitempty"`
 	// torrents config
 	Torrents TorrentsConfig `yaml:"torrents,flow,omitempty"`
+	// upload config
+	Upload UploadConfig `json:"scraper" yaml:"upload,flow,omitempty"`
```
akuma06: here json:"upload"

Kiloutre:
> **If you're not logged in**: **you only show the input text for the api key** and not the checkbox. If someone wants to upload to nyaasi/anidex, he will just need to fill the input.

What for? The idea is to allow anyone to upload to anidex & nyaa.si very easily from pantsu, merely checking a checkbox when you don't want to have to go through the bother of getting an anidex & nyaa.si API key is massively better
Anidex's account creation doesn't work at all for me and nyaa.si are too incompetent to show your API key somewhere in your settings, yet alone having an apidoc. Well great, now this whole PR is completely useless for me

Furthermore you're making it so that absolutely every single anonymous upload gets uploaded to anidex nyaa.si, again what is the point? We should upload the torrent to those websites only if the user asks to do so, either with him giving his API key or by having our own used when asked, the only thing this will accomplish is get our pantsu account on anidex & nyaa.si account banned from those websites with the garbage we tend to get

I ask for a rollback

akuma06:
Already reverted for non logged user if you checked the commit just after this one

Kiloutre:
Alright, the checkbox doesn't seem to work though?

akuma06:
I think it would be better to have the upload to the three platform in a go routine or there will be timeout loading when someone do the multiupload.
Then you load the state with some AJAX or manual F5.

akuma06:
It works but it seems there is a problem with how anidex detects their tracker in our torrent file.
For example, I tested to send the torrent file generated through nyaa to anidex (with anidex tracker in it, I checked in my bitorrent client) and it gives me an error "Anidex tracker not included".
I tested through http://torrenteditor.com/ to check again that there was anidex tracker. It was. Then I downloaded from their website (without any changes) the torrent file and uploaded it to anidex. And it worked.
So I don't know what might be wrong. If someone can test it to see if there is the same problem

Kiloutre:
Maybe it has to do with tracker order?
Looking at anidex torrent files, anidex is always at the very start of the trackers. Furthermore in their listing page, any tracker beside anidex is omitted, so i wouldn't be surprised if they only checked the very first tracker to see if it's the anidex one

Kiloutre:
I'm getting a 404 error when trying to upload to other websites, the torrent is uploaded & torrent file generated but /upload/status/:id gives me a 404
I also get this error message in the console
`[GIN-debug] [WARNING] Headers were already written. Wanted to override status code 302 with 500`

Torrent generation also doesn't seem to work during a regular download.go call, it says it's generating but it's seemingly not

akuma06:
Ok I will look into it when I have time :)

akuma06:
So I've modified some parts to the torrent generation when uploading torrent file.
The download part works and the upload to anidex too.

Now we need to only fill in nyaasi/tosho form and request part.

akuma06:
So I was looking to implement nyaasi upload but it seems that doing that would make less secure their credentials since nyaasi doesn't use an api key it seems.
I didn't find a documentation on how to use their api but I did find a python script (https://github.com/nyaadevs/nyaa/blob/master/utils/api_uploader_v2.py). It looks like you have to use a basic http authentification to use their api.
Beside the fact that asking the credentials from nyaasi to allow upload is not secure, I couldn't get their api to work with my Rest Client (same with their python file). It is always sending me this error: 
```
{
  "errors": {
    "recaptcha": [
      "The response parameter is missing."
    ]
  }
}
```

If someone can tell me if their python script is up to date and how their api works, that would be awesome!

ewhal (approved):

ewhal:
<3

PR #1644, opened 2017-10-08
"No API token" message if no API token, small changes

Kiloutre:
API token could be empty at times so it looked off in the user edit form to have the empty space
- " No API token" message if no API token
- Move "delete user" button to the right of the box, reducing the chances of a missclick
- Prettier tomorrow color on hovered button that actually are \<a\>

### DIFF
--- a/public/css/themes/tomorrow.css
+++ b/public/css/themes/tomorrow.css
@@ -115,6 +115,10 @@ td.tr-le, .error-text {
     background: hsla(40, 50%, 50%, 0.2) !important;
 }
 
+a.form-input[class*="btn"]:hover {
+    color: #dedede;
+}
+
 .pagination .active {
     background: #141517;
     color: #c5c8c6;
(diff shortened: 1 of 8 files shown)


Kiloutre:
@akuma06 I'm trying to get access of the T("") function from within template_function.go, in order to have formatThemeName(name string) return a translation string if it exists (e.g. themes_classic returning "nyaa.se (Beta)"), however after multiple tries i'm not getting any success
Any idea as to how do i proceed?

akuma06:
Why do you need that ? Can't you just use the translation directly ?
I'm asking this because using directly the translation function would mean initializing it again. It's not very efficient. That's why I would like to know what do you want to do exactly.

Kiloutre:
I wanna be able to have translation names "harcoded" into en_us.json for quick addition or deletion (or translation too), while also having a function that will format the name should no translation string exists (e.g. "**g** becomes **/g/**", because no one would do a one letter theme for any other reason)
It's not like i can rely on the automatic naming when i have theme names like "nyaa.se (Beta)" so i thought i would use the translation function to check if "themes_classic", "themes_g" or such exists and if it does return the translation

akuma06:
You can do the same as https://github.com/NyaaPantsu/nyaa/blob/10a70e16c9f8b22eef1adb39b1e20f985c678f52/templates/template_functions.go#L242
You just pass as an argument the T function and the string.

yiiTT (approved):

PR #1646, opened 2017-10-10
bottom margin for comment avatar & other small things

Kiloutre:
- Bottom margin for comment avatar, preventing text from being right under it
- Resetting API key redirects you on API key page (user edit) instead of user profile
- Move important faq informations up
- Fix modtools showing behind ads
- In upload form, if you clicked on upload without choosing a category, your browser would scroll you back to it but the select would be hidden by the header, this has now been fixed for JS users
- Fix "header-form" id accidentally used on two forms
- Fix very very slightly off-center input
- Fix HTML error in filelist
- Dates are now user-friendly constantly, even for non-JS users. JS is just used to convert to local time format

### DIFF
--- a/controllers/user/profile.go
+++ b/controllers/user/profile.go
@@ -235,5 +235,5 @@ func UserAPIKeyResetHandler(c *gin.Context) {
 	} else {
 		messages.AddInfoT("infos", "profile_updated")
 	}
-	UserProfileHandler(c)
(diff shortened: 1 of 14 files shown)


yiiTT (approved):
It should work right?

Kiloutre:
it WILL!

yiiTT:
travis will be the judge of that!

PR #1648, opened 2017-10-11
Make golang theme toggler behave exactly like JS one & css fix

Kiloutre:
- Make golang theme toggler behave exactly like JS one
- Make use of newly created default theme & default dark theme variables
- Fix red buttons on sukebei that had red text making it invisible


sfan5 (approved):
meme

PR #1651, opened 2017-10-11
Fix a condition that didn't behave as intended

Kiloutre:
List of things that i fucked up in this PR:

- User deletion (both through modpanel and user panel)

### DIFF
--- a/controllers/user/profile.go
+++ b/controllers/user/profile.go
@@ -22,13 +22,30 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+// UserProfileDelete :  Deleting User Profile
+func UserProfileDelete(c *gin.Context) {
+	id, _ := strconv.ParseUint(c.Param("id"), 10, 32)
+	
+	userProfile, _, errorUser := users.FindForAdmin(uint(id))
+	if errorUser == nil{
+		currentUser := router.GetUser(c)
+		if (currentUser.CurrentOrAdmin(userProfile.ID)) {
+			_, err := userProfile.Delete(currentUser)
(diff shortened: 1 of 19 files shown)


controllers/user/profile.go
```diff
@@ -28,7 +28,7 @@ func UserProfileHandler(c *gin.Context) {
 	Ts, _ := publicSettings.GetTfuncAndLanguageFromRequest(c)
 	messages := msg.GetMessages(c)
 
-	if id == 0 && ContainsNonNumbersChars(c.Param("id")) {
+	if c.Param("id") != "0" && id == 0 && ContainsNonNumbersChars(c.Param("id")) {
```
akuma06, line 48: If c.Param("id") is not "0" and not a number, then id = 0. So you don't have to check for ContainsNonNumbersChars.
The reason is simple, if you convert something else than a number to int, the variable is equal to 0.
So id is equal to 0 if c.Param("id") = "0" or if c.Param("id") is not a number

Kiloutre: The point of this condition is to send users to username search if the write /user/xxx , (which will make id into 0), but user/0 is the anonymous profile (also id 0) so i also have to check if it contains letters

akuma06: I understand :) and I think if you only do this `if c.Param("id") != "0" && id == 0` instead of this `if c.Param("id") != "0" && id == 0 && ContainsNonNumbersChars(c.Param("id"))`, you should get the same result. (explanations above)

yiiTT (approved):

PR #1662, opened 2017-10-14
Fix User delete button

### LINKED ISSUE #1652: Cannot delete user (self or other)
gives an invalid crsf token message and doesn't do shit

akuma06:
Fix #1652
This bug comes from the fact that @kiloutre edited how the form behave and instead of a GET request, do a POST one. However every POST request not in /api or /mod need a CSRF Token to work.


Kiloutre (approved):
Yeah i had to turn it into a post because otherwise users could include images that featured the user delete URL as the src in comments & description

PR #1674, opened 2017-10-16
[API] Corrected wrong method on profile

Kylart:
After trying I can tell this answers to `GET` only.

I hope this is the right place to modify. Please feel free to tell me if I need to change more than that.

### DIFF
--- a/apidoc/api_data.js
+++ b/apidoc/api_data.js
@@ -1044,7 +1044,7 @@ define({
       }
     },
     {
-      "type": "post",
+      "type": "get",
       "url": "/profile/",
       "title": "Get a user profile",
       "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


Kiloutre:
Could you tell me what doesn't work and how do i test it? I don't touch API stuff and the guy that does ( @akuma06 ) isn't available really often

Kylart:
This change only changes the documentation.

The documentation states that that route (`/api/profile`) answers to a `POST` call when in reality it only answers to `GET` calls.

How to test it though:
* Send a `POST` request to the said route with an ID of your choosing as data only to see that you'll get a `404 code`.
* Send a `GET` request to the said route with an ID of your choosing as parameters and you'll get the right result.

Kiloutre (approved):

PR #1684, opened 2017-10-18
Fixes & improvements

Kiloutre:
- Fix torrent auto-update that broke all torrent dates
- Show user status (Uploader, Banned or Moderator, in that priority order) in torrent comments
- Fix overly big bottom margin on last line of a comment
- Allow possibility of sending comments as Anonymous for logged-in users
- Fix overly big bottom padding for comments in classic theme
- Old nav enabled by default
- Fix a translation string that did not work in view.jet.jtml
- Possibly fix torrent dates that were 2h ahead of time
- Fix overly big spacing of categories in old navigation when low category count (Sukebei only has two)
- On-demand stat scraping frequency set in config file with limits for manual call of the script now existing, also set in config file


templates/template_functions.go
```diff
@@ -452,3 +454,28 @@ func formatDate(Date time.Time, short bool) string {
 		return fmt.Sprintf("%d/%d/%d, %d:%.2d:%.2d AM UTC+0", Date.Month(), Date.Day(), Date.Year(), Date.Hour(), Date.Minute(), Date.Second())
 	}
 }
+
+func getUserStatus(userid uint, uploaderID uint, commentList bool, T publicSettings.TemplateTfunc) template.HTML {
+	
+	user, _, errorUser := users.FindForAdmin(uint(userid))
```
Kiloutre: @akuma06 , this specific line is giving me a panic error for some reason. I tried changing it to FindByID, remove everything else besides this line, or put arguments in the function call that had no possible way of creating an error but no matter what as long as this line is here i'll get a panic error when the function is called
what do?

akuma06: First you shouldn't need uint cast since userid is already uint.
Second, you should use FindForAdmin only for some specifics page. Like mod panel not for what you want to do. Since you only want to get the user status, use FindRawByID()
Thirdly, it is not working because you shouldn't run sql queries from the template file. Every sql queries should be run before. For example, when generating the CommentJSON list, you have already access to User model. You should add the user status to the CommentJSON struct and fill it in the ToJSON function of Torrent struct.

sfan5 (approved):
desu

PR #1698, opened 2017-10-28
Redesign of notification page, lots of fixes & lots of improvements

Kiloutre:
**New features:**
- Full redesign of notification page, looks much better and clearer
- Added button to clear only already read notifications
- Added button to mark every notifications as read
- Notifications are now shown with their creation date
- Notification count is now shown in header on user avatar if above 0
- Allow reports to have messages (60char max)
- Show number of followers and total uploaded torrents size on user profile (used to be there but had to be removed because it was showing an incorrect count)
- Added button for admins to nuke all torrent reports

**Improvements:**
- Users commenting on their own torrent do not trigger notifications any more
- If you put an unexisting username in refine, the search will show no result instead of acting like you didn't specify any user
- Prevent duplicate trackers during torrent creation
- Possibly fix JS theme toggler not working on the very first click for logged-in users, make the query for non-logged in users less heavy for both the user and server
- Visual cue for locked torrents
- Admins modifying an user profile can force an e-mail without requiring validation
- Disclose that comments support markdown
- Further small improvements toward spacing in comments
- Add max-width and reasonable max-height for pictures in comments
- Small changes to bring the classic theme closer to the nyaa.se's theme
- During user torrent search, Old nav shows "Browsing X's torrents (Y)", X being the username and Y being the amount of torrents related to X, just like nyaa.se
- During user torrent search, page title includes the username instead of the usual "Home"
- Allow anonymous users to do reports, make it so that regular users do not need captcha to report
- Add a default avatar (when user has no avatar) that can be changed by modifying local pantsu files, default avatar is different for sukebei
- Remove formatDateRFC() function that became useless when i made the formatDate() function
- Fix the template_test_function part of some functions that was nowhere near proper
- Rename three functions that had my name in it, remove duplicate functions
- Announcements duration is now set as hours and not days

**Fixes:**
- Fix notification count in header requiring another refresh to update when accessing the torrent the notification is from
- Fix report deletion that did not work period for admins
- Fix visual artifact on the "More" links in admin panel, caused by the fact that the link was part of an html list
- Fix translation strings in admin panels that did not work because of extra spaces
- Fix unwanted background color on old navigation under sukebei's /g/ theme
- Fix file list indent not working for folders
- Fix some user settings requiring a second page refresh to update the changes in the form, _not all of them however_
- Fix wrong URL for lock button in admin panel
- Fix remake & anonymous button that didn't stay checked on upload fail
- Fix HTML error in old navigation and in torrent comments
- Fix announcement deletion
- Prevent user from following themselves, not that it do much
- Fix a translation string that was poorly translated (akumaaaaaa)


Kiloutre:
@akuma06 
How do i make it so that more recent notifications are shown first, just like torrents?

yiiTT (approved):

PR #1700, opened 2017-11-02
Make comment and users in admin panel ordered by newest to oldest

Kiloutre:
This will allow to look at newest users and comments from admin panel index without needing to go to the latest page everytime


yiiTT (approved):

yiiTT:
about time

Issue #1704, opened 2017-11-02
Spanish translation (WIP)


JuanjoSalvador:
Spanish translation is uncompleted (about 30% of sentences are translated) and has a lot of errors. I'm working on it.

If someone want to join me on the translation, please use my fork.

https://github.com/JuanjoSalvador/nyaa

JuanjoSalvador:
Status: WIP (99% translated)

**Unverified lines or needs a revision:**

ID: `large`
Reasson: Context needed

ID: `old`
Reasson: Context needed

ID: `delay`
Reasson: Context needed/Don't know how to translate

ID: `expire`
Reasson: Context needed/Don't know how to translate

Kiloutre:
> ID: large

Used in refine form: "Between 105MB and 2GB **large**"

> ID: old

It's not used anymore so don't worry about this one

> ID: delay

Not used anymore too

> ID: expire

Used in admin-panel, more precisely the announcements list
It shows at which precise time will the announcement expire
Pictured:
![pic](https://user-images.githubusercontent.com/11745692/32379447-d949fc38-c0ad-11e7-8192-0011bbf34675.png)

JuanjoSalvador:
Thanks @Kiloutre! I'm going to fix that then, 5 minutes.

JuanjoSalvador:
Fixed. Available to merge on #1707

PR #1707, opened 2017-11-03
Spanish translation (100% finished)

JuanjoSalvador:
Maybe needs some revisions/fixes by another user, but it's finished. See #1704 for more details.


Kiloutre (dismissed):

Kiloutre:
I've noticed that you didn't translate all category names, why is that?
pictures_photos is translated but software isn't translated fully (unless software works as a spanish word?), and art_doujinshi or real_life_photobooks_and_pictures are straight up not fully translated

Well it's alright because that was the case before anyway but it'd be nice to have those translated as well

JuanjoSalvador:
Wops, I didn't notice that. Let me fix those translations then before merge.

JuanjoSalvador:
OK, removed unused strings (goodbye `old` and `delay`) and fixes the translations of categories and the issue on `expire`.

Kiloutre (approved):

PR #1709, opened 2017-11-04
Add possibility of reporting without JS, add report reason for NSFW content, Add possibility of getting a torrent by Hash in refine

Kiloutre:
There was a /report/:id page but it wasn't being used, it now is
Added a "NSFW content" report reason
Add input to get a torrent by hash in refine and by the ?hash get parameter, though it only match exact hash and as such overrides any other search parameter

### DIFF
--- a/controllers/oauth/callback.go
+++ b/controllers/oauth/callback.go
@@ -80,7 +80,7 @@ func CallbackHandler(conf gooauth.Config) func(c *gin.Context) {
 			return
 		}
 
-		messages.AddInfo("infos", fmt.Sprintf("%s", token.Raw))
+		messages.AddInfo("infos", fmt.Sprintf("%s", token))
 		templateVariables := templates.Commonvariables(c)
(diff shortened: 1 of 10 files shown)


controllers/oauth/callback.go
```diff
@@ -80,7 +80,7 @@ func CallbackHandler(conf gooauth.Config) func(c *gin.Context) {
 			return
 		}
 
-		messages.AddInfo("infos", fmt.Sprintf("%s", token))
+		messages.AddInfo("infos", fmt.Sprintf("%s", token.Raw))
```
Kiloutre: @akuma06 this line, alone other lines using wrong values during string formatting, started creating errors a few hours ago on every PR
token itself is not a string but features a .Raw that is a string (and also a .Signature, along a few other variables that are not string)
Is that the correct thing to do, showing the raw token? I know nothing about oauth and how it works so i'm unsure if that's the right thing to do

yiiTT (approved):

PR #1715, opened 2017-11-06
Small fixes

### LINKED ISSUE #1716: Torrent generation not working on live because port is already in use
Port 50007 (default for anacrolix torrent client) is already in use so the torrent generation is not working on live

> \<ewhal4\> so you probably need to randomize it or make a queue for torrent file generation or something

Kiloutre:
- Subscribe link on classic theme is forced to be lowercase
- When an user saves his settings, it now cleans up old cookies of the current domain to avoid having duplicate cookies
- Add some further cookie cleanup to resetCookies() function
- Fix css rule that didn't apply to the wanted elements, causing text going outside of comments when comment as 500px or higher and images not to have size limits in comments
- Fix "username" input not working in admin panel's comment list
- Port for torrent client used by torrent generation can now be set through config file, should fix #1716 
- User notifications now sorted by newest (at the top) to oldest (bottom)
- Fix stat scraper not using given trackers if the total of tracker was under 4, although it has not happened a single time because a torrent always has at least 13 trackers, but it fixes unintended behavior
- Fix torrent reports done on non-default table name not showing in admin panel

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -106,7 +106,7 @@ torrents:
 # TorrentFileStorage : Path to default torrent storage location (eg /var/www/wherever/you/want)
 torrents:
 # GenerationClientPort : Port used by the torrent client created during torrent generation
-  generation_client_port: 50007
+  generation_client_port: 50006
 # FileStorage : Location of folder that will contain generated torrent files
   filestorage: ./downloads/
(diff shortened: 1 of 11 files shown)


yiiTT:
why do you need help?

Kiloutre:
The command "go get github.com/anacrolix/torrent" failed and exited with 1 during .

yiiTT (approved):

PR #1719, opened 2017-11-07
[WIP] On-demand filelist fetching

Kiloutre:
It's not fully completed (the JS that updates filelist doesnt show icons & no padding for stuff within folders) but it's fine for now and a mere refresh shows a proper filelist
Also it's very slow for long filelists right now, and outputs empty filename if there's only one file


sfan5 (dismissed):

controllers/torrent/files.go
```diff
@@ -0,0 +1,78 @@
...
+func ScrapeFiles(magnet string, torrent *models.Torrent, currentStats models.Scrape, statsExists bool) (error, []FileJSON) {
+	if client == nil {
+		err := initClient()
+		if err != nil {
+			return err, []FileJSON{}
+		}
+	}
+	
+	t, _ := client.AddMagnet(magnet)
+	<-t.GotInfo()
```
akuma06, line 48: If it's only for the filelist, you should detect before if you have the torrent file and then if you have it already, load the files from it (since you won't have to wait to connect to peers to have them). You can check how I access to a torrent file information in the upload process here with `tfile` the content of the torrent file 
https://github.com/NyaaPantsu/nyaa/blob/860f9c8eb1c001546402bd9a36c0fda9d97feb00/utils/validator/torrent/functions.go#L195

In the opposite, if you don't have it, you should save it after the scrapping so you save ressource power

akuma06, line 48: It might also be better to make it in a `go func()` when you try to scrap. So you don't keep the connection between the user and the server hanging. It might a while before yhe scrapper connects to a peer and gets the information.

Kiloutre: That itself is not a problem, since the torrent page load is not affected by this. The filelist fetching is generally done with JS, if the user doesn't have JS then a /files/:torrentid page is loaded but the user will known full well that if that page takes time to load, it's because it's loading the filelist
So it's not really necessary here

PR #1728, opened 2017-11-09
Janitor

Kiloutre:
Was checking if a variable was set but was checking an undefined variable

Moderator stuff:
- Add Janitors
- Janitors can lock torrents, delete comments and see/delete reports, ban users
- Remove any usage of deprecated HasAdmin() function
- Show lock/unlock buttons next to delete button in admin panel for torrents, and ban/unban button for users
- Add user ban route
- Banned users can still log-in, still upload, but cannot comment and their torrents are locked (invisible) by default
- Add page for moderation guidelines
- Moderators can edit the status of a torrent to undelete it

Other:
- Make locked torrents hidden from torrent listing for non-owners or non-staff
- Show every userstatus (Moderator, Janitor, Member, Banned, Trusted, a+) in comments instead of just Moderator, Banned, Trusted and Uploader
- No status on the anonymous profile
- Visual cue for locked torrents in admin panel
- Add current search (search content, or if empty search category, or if empty "Home") in page title
- Render torrent view template directly instead of redirecting user after submitting a comment
- Make notification page responsive
- Fix "torrent is being generated" message showing up even when the torrent couldn't be generated
- Fix the last html error caused by old nav
- Fix user profile that needed a refresh to visually update after a change
- Fix userprofile buttons overflowing at some specific resolutions
- Fix old (nyaa.se) comments not showing up in torrent listing
- Fix old (nyaa.se) comments's username that constantly showed as anonymous since a few weeks ago
- Fix on-demand stat fetching that tried to scrape HTTP trackers as they were in the default trackers, which is useless because it can only scrape UDP scrapers

### DIFF
--- a/controllers/activities/activities.go
+++ b/controllers/activities/activities.go
@@ -32,7 +32,7 @@ func ActivityListHandler(c *gin.Context) {
 	}
 	var conditions []string
 	var values []interface{}
-	if userid != "" && currentUser.HasAdmin() {
+	if userid != "" && currentUser.IsModerator() {
 		conditions = append(conditions, "user_id = ?")
 		values = append(values, userid)
 	}
--- a/controllers/api/api.go
+++ b/controllers/api/api.go
@@ -433,7 +433,7 @@ func APISearchHandler(c *gin.Context) {
 		userID = 0
 	}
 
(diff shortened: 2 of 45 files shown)


yiiTT (approved):
would be nice to un-delete torrents and list deleted torrents somewhere

Kiloutre:
That's already the case m8

PR #1749, opened 2018-01-04
Fix parseAllDates() that could break dates (in listing)

Kiloutre:
Fix parseAllDates() that would break torrent list dates if it had already been executed before

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -59,8 +59,10 @@ function parseAllDates() {
   var list = document.getElementsByClassName("date-short")
   for(var i = 0; i < list.length; i++) {
     var e = list[i]
-    e.innerText = new Date(e.title).toLocaleString(lang, ymdOpt)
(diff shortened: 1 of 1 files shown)


public/js/main.js
```diff
@@ -59,8 +59,10 @@ function parseAllDates() {
   var list = document.getElementsByClassName("date-short")
   for(var i = 0; i < list.length; i++) {
     var e = list[i]
-    e.innerText = new Date(e.title).toLocaleString(lang, ymdOpt)
-    e.title = new Date(e.title).toLocaleString(lang)
+    var newDate = new Date(e.title)
+    if(newDate == "Invalid Date") continue
```
akuma06: `if (isNaN(newDate.valueOf()) continue` would do a better job, since a text error can easily change between javascript engine or browser.

ewhal (approved):

PR #1753, opened 2018-01-08
Allow class "new" to be applied to header text

Kiloutre:
pic

### DIFF
--- a/public/css/main.css
+++ b/public/css/main.css
@@ -726,12 +726,11 @@ textarea {
 .wew:after { 
     content: "wew";
 }
-#header .new:after, .header .wew:after
+#header .new:after, .header .wew:after {
     font-size: 7pt;
     vertical-align: middle;
     position: relative;
     top: -8px;
-}
 }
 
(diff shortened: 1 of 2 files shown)


public/css/main.css
```diff
@@ -715,15 +715,23 @@ textarea {
...
     color: red;
-    font-size: 10pt;
+    font-size: 8pt;
     vertical-align: top;
     line-height: 5px;
     margin-left: 1px;
 }
-#commit.wew:after { content: "wew"; }
+.wew:after { 
+    content: "wew";
```
akuma06, line 727: Shouldn't this be "new" instead of "wew"

Kiloutre: This is some old thing that i put because at time the commit hash would show "unknown", and as such i would show "wew" on it instead of "new"
It can be removed

akuma06: Oh ok, no keep it if you want. I thought it was a typo

ewhal (approved):

PR #1770, opened 2019-06-14
translating zh-cn & make some fix 

ted423:


ewhal:
good to go?

ted423:
> good to go?

done.
But I still have some doubts

en-us has duplicate ID "error_len_array"
one in 1635 row and one is 1719,could I delete one ?


in row 1715 ,there is a id  "error_"  i think is a wrong copy of ”error_len_array“(row 1719)

PR #1779, opened 2021-06-01
Add Hindi translation

Saurmandal:


translations/is-is.all.json
```diff
@@ -1350,6 +1350,9 @@
   {
     "id": "language_fr-fr_name",
     "translation": "French"
+  },
+    "id": "language_hi-in_name",
```
akuma06, line 1355: Missing an open bracket here :)

akuma06 (changes requested):

translations/is-is.all.json
Saurmandal: Fixed, thank you :)

akuma06 (approved):

